PR #495, opened 2013-05-31
Float tab bar at the bottom of the screen.

xiongchiamiov:
When pagination is set to anything more than, say, 10, it's kinda
annoying to have to scroll down to the bottom of the page to do anything
with an error.  This is particularly the case if you're doing things to
the first couple errors, which is what I'm usually looking at.

Now the actions are always at hand.


stevecrozz:
Closing this due to inactivity

PR #498, opened 2013-06-06
Canonicalize problem notice urls

danielbeardsley:
Previously the notice visible at `/apps/:id/problems/:id?notice=X`
would change each time a new notice was reported because pagination was
using reverse ordering (`notice=1` is most recent). This made it
problematic to show someone a particular notice via a URL. For
any given url, the shown notice changed each time a notice was
reported.

Now, notices are indexed such that ?notice=1 will show the oldest
notice. This makes it so the content avaiable at that URL doesn't keep
changing every time a notice is reported.

Leaving off the ?notice=X param still defaults to the most recent
notice but the "Older" and "Newer" links still work as you'd expect.

This change simplifies the code, drops a (now-unused) helper method, and
makes errbit less surprising.

Note: the prev/next templates now make more logical sense with `prev == Older`.


rud:
Going to close this out as the code is very outdated. Feel free to revive the idea in the future with a fresh PR.

PR #574, opened 2013-10-08
update notifier.js of printstacktrace

ctokoro:
Hi! I use javascript notifier, but backtrace was not working well.

i check inside of notifier.js , i found out it caused by printstacktrace.
i update that code from stacktracejs.com and it seems to work better.(but not perfect yet)
( http://cloud.github.com/downloads/eriwen/javascript-stacktrace/stacktrace-0.4.js )

p.s.
stacktracejs's newest version is v0.5.3 ( https://github.com/eriwen/javascript-stacktrace ) but it seems unfit to errbit notifier.(maybe it should fix another code.)


stevecrozz:
We're planning going to phase out notifier.js in favor of airbrake-js. See #701

PR #631, opened 2013-12-26
Applications list is not ordered by problems count

alvarobp:
After upgrading to an errbit version using Mongoid v3 we've observed that the applications list does not show up the applications ordered by the number of unresolved problems count. Since we have a considerable number of different applications playing together we were relying heavily on that list being ordered.

The issue comes from Mongoid v3 defining a ``sort`` method on the Criteria class masking the sort method defined in Enumerable, thus the ``<=>`` method is not used when sorting which is the one implementing the sort by unresolved problems count.

This suggested patch reverts only to sorting using the full array of App instances, which is the previous implementation.

Do you think it makes sense to revert to the previous behaviour?

### DIFF
--- a/app/controllers/apps_controller.rb
+++ b/app/controllers/apps_controller.rb
@@ -12,7 +12,7 @@ class AppsController < ApplicationController
   }
 
   expose(:apps) {
-    app_scope.all.sort.to_a
+    app_scope.all.to_a.sort
   }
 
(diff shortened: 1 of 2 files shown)


stevecrozz:
It looks we fixed this in another commit:
https://github.com/errbit/errbit/blob/master/spec/controllers/apps_controller_spec.rb#L46

PR #632, opened 2014-01-01
Instruction to deploy on OpenShift

parallel588:
Hi all, 
I added instructions to deploy on OpenShift.
What do you think about it?

Thanks.


stevecrozz:
It's very cool, but I don't want to have to support this deployment method because I have no experience with OpenShift. I do appreciate having this information available though. I would be happy to link from the documentation to your own blog if you have this information up somewhere. If you wanted to do that, I would add it here next to the systemd config link:
https://github.com/errbit/errbit/blob/master/docs/deployment.md

PR #659, opened 2014-02-19
Advancedsearch

deathowl:
Advanced search feature added to separate branch, now it's mergeable

### DIFF
--- a/app/assets/javascripts/form.js
+++ b/app/assets/javascripts/form.js
@@ -21,12 +21,20 @@ $(function(){
 });
 
 function activateAdvancedSearchAPI() {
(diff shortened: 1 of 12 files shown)


stevecrozz:
This didn't make milestone 0.3.0, but I think we may want to explore some other advanced search ideas, including different syntax and search engines before moving any further with this. Hop onto #errbit or the mailing list if you want to discuss some plans.

arthurnn:
I agree. at Shopify we are using ElasticSearch with errbit as search engine. Errbit should provide a way to plug different search engines.. lets discuss that further.

PR #664, opened 2014-02-28
Filter cleaner

deathowl:
A CLI tool to search and delete errors from command line matching a query.
Example:
root@bud-errbit01:/opt/errbit# rake 'errbit:db:clear_filtered[a, 1, false]'
Errors returned by the Query:
Demo App 907422662/FooError: Too Much Bar
Destroying problem
=== Cleared 1  matching errors from the database.
root@bud-errbit01:/opt/errbit# rake 'errbit:db:clear_filtered[a, 1, true]'
Errors returned by the Query:
Demo App 587008510/wrong number of arguments (3 for 0)
=== No errors were hurt during the task

### DIFF
--- a/app/interactors/filter_cleaner.rb
+++ b/app/interactors/filter_cleaner.rb
@@ -6,11 +6,13 @@ class FilteredProblemClearer
   # Clear all problem already resolved
   #
   def execute(args)
+    puts("Errors returned by the Query:")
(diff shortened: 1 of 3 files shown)


rud:
Thank you for this contribution. To me the Rails console seems like an easier way to perform these one-off cleanups, what do you think?

I'm going to close this PR, hope you don't mind.

PR #709, opened 2014-07-09
Use Github Oauth, instead of username and password.

sinsoku:


stevecrozz:
This feature seems to be implemented already in the github plugin:
https://github.com/errbit/errbit_github_plugin

PR #715, opened 2014-08-10
Apps and users list sorting

Strnadj:
Same problem sorting but in app and user list.


rud:
This is a pretty neat implementation. Would you be willing to rebase it on master, @Strnadj? Thanks in advance!

stevecrozz:
Closing this due to age

PR #716, opened 2014-08-13
FIX inline comments styles

oelmekki:
On error list page, comments with especially long words (typically :
urls) break layout.

errbit

Fixed it by using the same forced width and wrapping settings than on
error message.


arthurnn:
thanks

PR #730, opened 2014-09-17
Prowl notifications

boncey:
Have added Prowl notifications to errbit.
http://www.prowlapp.com


rud:
Thank you for this contribution. I think this is outside of the scope of the project, so I move to close this PR.

boncey:
I don't see how it's any more or less in scope than the other notifications.

Saying that, my fork suits my needs so I don't really mind if it's accepted or not.
No idea if anyone else finds it useful.

stevecrozz:
Thanks @boncey. I do appreciate the contribution and you're right that this is much like the other notification systems. The other notifiers have been kinda "grandfathered" in, under an older approach where we were including everything. Maybe someday we'll finally get our plugin-based notifiers, but for now I'm glad you've got things working on your fork.

PR #731, opened 2014-09-18
Fix pivotal tracker integration removing requested_by attribute

jpl:
The "requested by" attribute needs to match username in pivotal tracker (which is often not the case, and seems quite buggy on the pivotal side), and is not mandatory as the request is by default owned by the token api submitted during integration.


stevecrozz:
All the issue tracker plugins are in individual repos now. The pivotal plugin currently lives at:
https://github.com/brandedcrate/errbit_pivotal_plugin

If anyone makes a few good contributions and wants to own the repo, I'd be happy to transfer it.

PR #749, opened 2014-11-06
Add date filter for problems/search

robinboening:
Added two input fields with a javascript datepicker to problems/search form to filter problems by date range. first_notice_at and last_notice_at are both considered when searching:
- Problem which was first noticed at 2014-01-05 will be found if your search criteria is for example 2014-01-01 until 2014-02-01.
- Problem which was last noticed at 2014-01-05 but first noticed at 2013-12-24 will also be found if your criteria is for example 2014-01-01 until 2014-02-01.

By default the very first notice date of the very first problem is used as the beginning value of the date range. The end value is by default the date of tomorrow. So searching for something but not a date will not lead to wrong results.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -125,5 +125,6 @@ group :assets do
   gem 'jquery-rails', '~> 2.1.4'
   gem 'pjax_rails'
   gem 'underscore-rails'
+  gem 'pickadate-rails'
   gem 'turbo-sprockets-rails3'
 end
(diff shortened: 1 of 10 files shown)


rud:
This seems like a good and sound change. Would you be willing to rebase/reimplement it on master @robinboening ? If not I think we should close this PR.

robinboening:
Hey @rud sorry I haven't seen your previous answer. Yes I can rebase with current master. Can you open the issue again please?

Cheers
Robin

rud:
@robinboening feel free to create a new PR, it's easier :)

robinboening:
OK. I just noticed the internal API changed a lot since 2014. It will take more time than expected to update my code.

robinboening:
Link to new PR https://github.com/errbit/errbit/pull/1448

PR #753, opened 2014-11-12
iMessage notification service

otzy007:
It works only under OS X.


rud:
This does not make running the service in production easier as it is limited to OS X. Going to close this for now, hope that's okay.

Issue #760, opened 2014-11-19
unicorn.rb should be an example file


simmerz:
Unicorn configuration, just as with mongoid and others, should be an example-only configuration, allowing easier management of real-world configuration files.

This will allow those with (for example) git deploy strategies to not have to alter their strategy to handle the unicorn.rb file imposed in errbit.

stevecrozz:
Closing this, explanation in #761, fixed in #777

PR #761, opened 2014-11-20
Move unicorn.rb out of the way, allowing for custom deploy strategies

simmerz:
Fixes the issue raised at https://github.com/errbit/errbit/issues/760


arthurnn:
I was thinking about this, and TBH, I think we need to jump way too many hoops to support heroku out of the box. i.e. https://github.com/errbit/errbit/blob/master/config/initializers/mongo.rb , https://github.com/errbit/errbit/blob/master/config/initializers/_load_config.rb .

I think we should have mongoid.yml, deploy.rb, unicorn.rb files (config files) , all as example, and let `errbit:copy_configs` copy them. People using chef/puppet, can override those, and people on heroku would have to change the .gitignore file and push those to heroku. 
Like that we would have to care about less config, and we would be able to please everybody..
@stevecrozz thoughts?

stevecrozz:
I think these hoops are mainly about supporting two methods of configuring errbit, but the one I prefer is actually the one we're using for heroku deploys (12-factor style environment variables). I think things are working out pretty well in the current state of affairs. Maintaining heroku support doesn't seem like too much trouble.

Although I think there's still something we can do here. For instance, I've been thinking of switching to puma anyway and there's a good chance we wouldn't need to package errbit any special puma config.

arthurnn:
Sure I like the ENV configs. What I dont like is we having to write initializers to change mongoid configurations. Theorically the app should always have a mongoid.yml file. which could use ENV vars.

That said, people would just have to run the copy task and push that to heroku

stevecrozz:
We should probably discuss general configuration guidelines a bit, but I don't think we should drop out-of-box heroku deployment support to get better puppet support because it should be relatively easy to solve this ticket and continue to support heroku the way we always have with something like this:
https://github.com/errbit/errbit/pull/777

@simmerz I'd also like to know what kind of custom configuration you need to apply and if it makes sense to support whatever you're doing by adding more knobs to tweak via environment variables.

stevecrozz:
I'm closing this in favor of #777 which is almost the same fix as the one you provided @simmerz, but it preserves heroku-compatible deploys.

PR #766, opened 2014-12-05
Modify Slack to use WebHooks and fix HipChat url

jerrett:
This is a breaking change, but it seems it's the right way to integrate with slack for this.. 

Using the WebHooks it means people can customize the name used to post as well as the icon. This also creates a much simpler setup, and one that looks to be supported. The current setup sort of works, but it seems to be using an old deprecated url format in Slack, and it seems to be using their API key for a WebHook call. I suspect things were setup different @ slack when this code was written!

I also dramatically improved the display of the message that gets sent, and fixed the problem_url for HipChat (currently if you set the protocol to https, the HipChat notification ignores it and you get http links anyway.)

### DIFF
--- a/app/models/notification_services/hipchat_service.rb
+++ b/app/models/notification_services/hipchat_service.rb
@@ -22,7 +22,7 @@ if defined? HipChat
     end
 
     def create_notification(problem)
-      url = app_problem_url problem.app, problem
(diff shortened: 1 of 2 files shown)


rud:
Thank you for this change @jerrett  - would you be willing to rebase it on master?

PR #777, opened 2015-01-02
move unicorn.rb to unicorn.default.rb

stevecrozz:


TeatroIO:
I've prepared a stage. <a href='http://default-unicorn-config.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com' target='_blank'>Click to open.</a>

nofxx:
No .gitignore for config/unicorn.rb ?

stevecrozz:
I see that other configs are in .gitignore, is that why you're asking? Or is there another good reason to do it?

I would imagine that someone who runs errbit via unicorn would fork this repo and check in a config/unicorn.rb and they wouldn't necessarily want it to be .gitignored.

nofxx:
Yup, to avoid confusion as all configs to be linked are there.
Should someone commit a config/unicorn.rb ?

stevecrozz:
It depends on how you deploy errbit. If you use heroku or foreman, you won't need to do anything as long as you're happy with config/unicorn.default.rb

But if you want some custom unicorn config, you'll have to create a config file somehow before you run unicorn. That could be by checking it into source control, or some other method. It doesn't need to be called config/unicorn.rb, but that would probably be a sane name to use.

arthurnn:
`config/unicorn.rb` should be on the gitignore, like `deploy.rb` is. So people can inject that file using chef or puppet, and we make sure we dont commit that file too.

arthurnn:
Done via https://github.com/errbit/errbit/commit/bd3eb276c0dbbf7661f548664e29b933b88641da

PR #778, opened 2015-01-02
update configs for rails 4.1

chulkilee:
- use secrets.yml and simplify secret_token configuration
- use Rails.application over Errbit::Application
- update other config with rails:update and blank apps


TeatroIO:
I've prepared a stage. <a href='http://chulkilee-rails41.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com' target='_blank'>Click to open.</a>

config/initializers/cookies_serializer.rb
```diff
@@ -0,0 +1,3 @@
+# Be sure to restart your server when you modify this file.
+
+Rails.application.config.action_dispatch.cookies_serializer = :marshal
```
arthurnn: why :marshal? I would set to :json as is the new version.

config/initializers/devise.rb
```diff
@@ -240,12 +251,17 @@
...
   # When using omniauth, Devise cannot automatically set Omniauth path,
   # so you need to do it manually. For the users scope, it would be:
-  # config.omniauth_path_prefix = "/my_engine/users/auth"
+  # config.omniauth_path_prefix = '/my_engine/users/auth'
 end
+
+# The secret key used by Devise. Devise uses this key to generate
+# random tokens. Changing this key will render invalid all existing
+# confirmation, reset password and unlock tokens in the database.
+Devise.secret_key = Rails.application.config.secret_token
```
arthurnn: I am pretty much sure this is not necessary after https://github.com/plataformatec/devise/pull/2835 . I am bumping devise installation in here so we will have that patch.

config/deploy.example.rb
```diff
@@ -59,7 +59,7 @@
     # Generate unique secret token
     run %Q{if [ ! -f #{shared_configs}/secret_token.rb ]; then
```
arthurnn: this line is not necessary anymore, as we are removing `secret_token.rb` in this PR.

arthurnn:
thanks for the help, eager to merge this after you work on my comments.

config/initializers/cookies_serializer.rb
chulkilee: I think the default behavior is marshal and I didn't want to change it to avoid breaking existing sessions.

- Setting it to json will break existing sessions after upgrade
- Using hybrid will break new (json) sessions if the server is rollbacked across this commit.

config/initializers/devise.rb
chulkilee: I'll make change accordingly when devise is upgraded.

arthurnn: Just rebase you branch, already upgrade Devise on master.

chulkilee: I don't think so - the current version of Devise in Gemfile.lock on master branch is 3.1.1. The fix is included from 3.2.0

stevecrozz: I went ahead and updated the devise gem to 3.4.1 on master

arthurnn: LoL, I had that locally but forgot to push... thanks @stevecrozz

chulkilee:
Hm.. there are several issues

- cookie serialization format: json, hybrid or marshal
- secret_token vs secret_key_base

It might be possible to keep the previous behavior (marshal / secret_token), but I think it's okay to move forward to keep it updated with Rails - use hybrid / secret_key_base.

I also believe errbit doesn't need to be "deploy-ready" project by itself. If some configuration should be changed for updated dependencies, users need to work with that. Of course errbit needs to provide some documentation.

What do you think about it?

arthurnn:
I think, we should always move forward, and avoid having 2 code paths as much as possible. Meaning, we need to use latest ruby/rails and use latest cookie serializer and secret_key_base.

Regarding deploy, if secrets.yml using a ENV var, that would make "deploy-ready" possible, as people can just set the variable.

config/initializers/cookies_serializer.rb
chulkilee: Changed it to hybrid to move forward :)

chulkilee:
Although the code works, let's merge it after merging #787. I had problem of compiling asset using capistrano since Devise requires secret_key which is set by env variable. I can do it manually with setting env variable, but having dotenv-deployment will handle both heroku and non-heroku deployment without additional step for setting them.

chulkilee:
I rebased dotenv-deployment commits to this branch since that is required to import .env in production.

Now this branch is ready to merge - I'm tested on my server. Do we need to other docs?

Gemfile
```diff
@@ -24,11 +24,13 @@ gem 'rails_autolink'
 # Please don't update hoptoad_notifier to airbrake.
 # It's for internal use only, and we monkeypatch certain methods
 gem 'hoptoad_notifier', "~> 2.4"
-gem 'draper', :require => false
+gem 'draper'
 
 gem 'errbit_plugin', github: 'errbit/errbit_plugin'
 gem 'errbit_github_plugin', github: 'errbit/errbit_github_plugin'
 
+gem 'dotenv-deployment'
```
arthurnn: so now that we are on capistrano 3, why do e need this gem?

chulkilee: dotenv-deployment is required to use .env in production.

Capistrano 3 has a built-in feature to link files (`linked_files`) so deploy.rb does not need to include dotenv-deployment capistrano task - but still dotenv-deployment should be included in Gemfile.

stevecrozz: I'm still thinking this over the dotenv stuff. In the long-term, I don't want to support multiple different configuration paradigms. ENV-based seems way more flexible than the sea-of-YAML approach. So that's a win for dotenv. But we should consider exactly how we want configuration to work. It seems like the only thing we really want out of dotenv-deployment is the one line:
```Dotenv.load('.env')```

It does do some other stuff, but do we want that other stuff?

If we move this direction are we going to provide a .env.example file or something similar? Is that going to be the place where *all* of the configuration goes? Is it going to be required to provide a .env file in development? in production?

We still have other ways of providing configuration (configuration files and passing in environment variables) from outside. What order of precedence should there be if you specify the same config as an ENV, in .env and in a yaml file?

chulkilee: 1) errbit is already using two configuration paradigms - config.yml and ENV (e.g. config/initializers/mongo.rb). Therefore adding dotenv-deployment does not make big changes in configuration methods in errbit - it justs add easy way to configure it when deploying.

2) dotenv-deployment is pretty minimalistic as you described - just load .env to ENV. It does not do any other magic. For example, if ENV is manually passed, values from .env will not overwrite it, so there is (and should be) no harmful side effect.

3) Also note that secret.yml has default value for secret_key_base, so you do not need to create .env file for development.

Without dotenv-deployment, I cannot use capistrano "out-of-box" since I'm not deploying to Heroku. That's the main reason I think it should be part of this pull request. Probably I should have made separate pull request prior to this one, but all things happened at the same time on my side - sorry about that :)

So my proposal is 1) merge this request with dotenv-deployment and then 2) simplify configuration in some way.

stevecrozz: I'm on board with this approach. I was just wondering if we want people to be able to do all this stuff:
https://github.com/bkeepers/dotenv-deployment/blob/master/lib/dotenv/deployment.rb#L8-L14

It seems pretty weird, but I guess its ok. I'll just not use that feature.

stevecrozz:
@arthurnn are you finished reviewing this?

arthurnn:
Will have a last look tomorrow.. thanks all

stevecrozz:
ping @arthurnn

arthurnn:
This is good... sorry about the late response. and thanks

PR #788, opened 2015-01-05
errbit unicorn systemd service

nofxx:
A heavily copied from gitlab systemd service, works nice. Optionally might add:

    #Restart=on-abort

As a comment maybe, it's like god for systemd, it monitors the process and restart it of failure/quit.

### DIFF
--- a/config/errbit-unicorn.service
+++ b/config/errbit-unicorn.service
@@ -1,0 +1,31 @@
+[Unit]
+Description=Errbit Unicorn Server
+Requires=mongodb.service
+After=mongodb.service network.target
+
+[Service]
+# Change here: user, group and directory to errbit root
+User=errbit
+Group=errbit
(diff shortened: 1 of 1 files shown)


TeatroIO:
I've prepared a stage. <a href='http://fireho-systemd.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com' target='_blank'>Click to open.</a>

stevecrozz:
I'm actually attempting to winnow down the number of things we have to support, so I can't accept this into errbit source. But what I would suggest is to create a public gist with this file and I'll update our documentation to point to your gist. That way we can offer some help for people using systemd without having to maintain that support ourselves.

If you send me a link to your public gist with this file, I'll update our docs to point to it.

nofxx:
Ok, here it is: https://gist.github.com/nofxx/f01dcfe3e9d504181d76
Thank you.

PR #790, opened 2015-01-05
Fix favicon URL

felixbuenemann:
Without this change the url points to /images/favicon.ico, which does not exist.


TeatroIO:
I've prepared a stage. <a href='http://felixbuenemann-fix-favicon.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com' target='_blank'>Click to open.</a>

arthurnn:
good catch

arthurnn:
But I wonder if we should actually move the favicon to that location, so sprockets can digest that file ?!?

felixbuenemann:
There are pros and cons. If you don't have a /favicon.ico you will get lot's of 404s from user agents that are trying to directly fetch the favicon without looking at the headers. On the other hand invalidating a non-digested favicon across browsers is very hard. If you want the best of both worlds, move the favicon to app/assets/images and symlink it to public.

If you prefer that way, I can push a new commit.

arthurnn:
good idea... lets do the symlink dance.

felixbuenemann:
OK, moved favicon to app/assets/images and added a symlink in public.

arthurnn:
thanks

PR #791, opened 2015-01-05
Fix double escaping of problem title in list view

felixbuenemann:
Without this change, the truncated text, which is already html escaped gets escaped again in the map block.

### DIFF
--- a/app/helpers/problems_helper.rb
+++ b/app/helpers/problems_helper.rb
@@ -6,7 +6,7 @@ module ProblemsHelper
   def truncated_problem_message(problem)
     unless (msg = problem.message).blank?
       # Truncate & insert invisible chars so that firefox can emulate 'word-wrap: break-word' CSS rule
-      truncate(msg, :length => 300).scan(/.{1,5}/).map { |s| h(s) }.join("&#8203;").html_safe
+      truncate(msg, :length => 300, :escape => false).scan(/.{1,5}/).map { |s| h(s) }.join("&#8203;").html_safe
     end
   end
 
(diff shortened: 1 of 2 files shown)


TeatroIO:
I've prepared a stage. <a href='http://felixbuenemann-fix-problem-title-double-escape.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com' target='_blank'>Click to open.</a>

shingara:
Have you an failure example ? This PR can introduce some security issue too :(

felixbuenemann:
@shingara how can it introduce a security issue? The text is always escaped in the map block using the `h` helper. So because truncate already addy html escaping, content is escaped twice.

shingara:
ok, you right.

arthurnn:
Can you add a test case for this? thanks

felixbuenemann:
OK, added a spec to check for double escaping that will fail *without* the changes:

```
Failures:

  1) ProblemsHelper#truncated_problem_message does not double escape html
     Failure/Error: expect(truncated).to_not include('&amp;')
       expected "#&amp;lt;&#8203;NoMet&#8203;hodEr&#8203;ror: &#8203;...&amp;g&#8203;t;" not to include "&amp;"
     # ./spec/helpers/problems_helper_spec.rb:16:in `block (3 levels) in <top (required)>'
```

You can test by setting `truncate(…, escape: true)` in truncated_problem_message.

shingara:
great thanks :+1:

arthurnn:
awesome. thanks!

PR #794, opened 2015-01-07
capistrano 3

chulkilee:
### Breaking changes
- No more deployment section in `config.yml`. Now need to customize `deploy.rb` and `production.rb`.

### Dropped features
- `errbit:setup_config` does not create secret_token.rb

### New features
- added capistrano-rbenv: enabled when `rbenv` env variable is set when running capistrano. (e.g. `rbenv=1 bundle exec cap production deploy:check`)

### DIFF
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -62,19 +62,6 @@ use_gravatar: true
 # Default Gravatar image, can be: mm, identicon, monsterid, wavatar, retro.
 gravatar_default: identicon
 
-# Setup your deploy options for capistrano.
-deployment:
-  hosts:
-    web: errbit.example.com
-    app: errbit.example.com
-    db: errbit.example.com
-  repository: https://github.com/errbit/errbit.git
-  branch: master
-  user: deploy
-  deploy_to: /var/www/apps/errbit
-  # setup path to unicorn pids folder (or deploy_to/shared/pids will be used)
-  # pids: /var/www/apps/errbit/shared/pids
-
 # GitHub OAuth configuration
 # If you want to allow authentication via GitHub, you will need to register
 # your app at: https://github.com/settings/applications
--- a/config/deploy.example.rb
+++ b/config/deploy.example.rb
@@ -6,98 +6,87 @@
 # `cap deploy` whenever you would like to deploy Errbit. Refer
 # to the Readme for more information.
 
-config = YAML.load_file('config/config.yml')['deployment'] || {}
+# config valid only for current version of Capistrano
+lock '3.3.5'
 
-require 'bundler/capistrano'
-load 'deploy/assets'
+set :application, 'errbit'
+set :repo_url, 'https://github.com/errbit/errbit.git'
+set :branch, ENV['branch'] || 'master'
+set :deploy_to, '/var/www/apps/errbit'
+set :keep_releases, 5
 
-set :application, "errbit"
-set :repository,  config['repository']
+set :pty, true
+set :ssh_options, forward_agent: true
 
-role :web, config['hosts']['web']
-role :app, config['hosts']['app']
-role :db,  config['hosts']['db'], :primary => true
+set :linked_files, fetch(:linked_files, []) + %w(
+  .env
+  config/config.yml
+  config/mongoid.yml
+  config/newrelic.yml
+  config/initializers/secret_token.rb
+)
 
-set :user, config['user']
-set :use_sudo, false
-if config.has_key?('ssh_key')
-  set :ssh_options,      { :forward_agent => true, :keys => [ config['ssh_key'] ] }
-else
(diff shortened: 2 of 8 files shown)


TeatroIO:
I've prepared a stage. <a href='http://chulkilee-capistrano3.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com' target='_blank'>Click to open.</a>

config/config.example.yml
```diff
@@ -64,13 +64,15 @@ gravatar_default: identicon
...
-    app: errbit.example.com
-    db: errbit.example.com
-  repository: https://github.com/errbit/errbit.git
+  servers:
+    errbit.example.com:
+      roles:
+        - app
+        - db
+        - web
+      user: deploy
```
stevecrozz: I can't see where the user param is used. Is it still configurable?

config/deploy.example.rb
```diff
@@ -6,98 +6,70 @@
...
-    "`cat #{path}/unicorn.pid`"
-  end
+pids_path = config['pids'] || "#{fetch(:deploy_to)}/shared/pids"
+set :unicorn_pid, "`cat #{pids_path}/unicorn.pid`"
 
+namespace :unicorn do
   desc 'Reload unicorn'
-  task :reload, :roles => :app, :except => { :no_release => true } do
-    run "kill -HUP #{unicorn_pid}"
+  task :reload do
```
stevecrozz: Seems like this should also be on roles(:app)

config/deploy.example.rb
```diff
@@ -6,98 +6,70 @@
...
-  end
+set :pty, true
+if config.key?('ssh_key')
+  set :ssh_options, forward_agent: true, keys: [ config['ssh_key'] ]
+else
+  set :ssh_options, forward_agent: true
 end
 
-namespace :errbit do
-  task :setup_configs do
```
stevecrozz: This task seems pretty useful. Why remove it?

config/config.example.yml
chulkilee: `roles` and `user`  will be passed to `server` in production.rb as `opt`.

Having several `server` with same name doesn't seem to work - that's why I changed a way to organize by server, not by role.

.capistrano/metrics.example
```diff
@@ -0,0 +1 @@
+none
```
stevecrozz: We should probably not check this file in so users can decide whether to send metrics.

config/deploy.example.rb
```diff
@@ -6,98 +6,70 @@
...
-set :repository,  config['repository']
-
-role :web, config['hosts']['web']
-role :app, config['hosts']['app']
-role :db,  config['hosts']['db'], :primary => true
-
-set :user, config['user']
-set :use_sudo, false
-if config.has_key?('ssh_key')
-  set :ssh_options,      { :forward_agent => true, :keys => [ config['ssh_key'] ] }
```
stevecrozz: Why remove the ssh_options?

.capistrano/metrics.example
chulkilee: This is example file.

config/deploy.example.rb
chulkilee: See the description - it is "WIP".

config/deploy.example.rb
chulkilee: done - thanks for the catch

config/deploy.example.rb
stevecrozz: Ah. I understand now. I forgot it was a WIP while I was reviewing.

stevecrozz:
@chulkilee This is looking really good. Thanks for your efforts so far.

config/deploy.example.rb
chulkilee: Added back except secret_token.rb creation.

I'm using rbenv and calling bundle exec inside script doesn't work very well. Once #778 is merged, we will not need to handle `secret_token.rb`, so I remove it.

.capistrano/metrics.example
arthurnn: can we just not checkin the example?

chulkilee: removed

config/config.example.yml
chulkilee: I removed the deployment section from config.yml. Having deployment configuration on config.yml is confusing since config.yml file should be for app configuration.

arthurnn:
I dont have any capistrano setup to test this. @chulkilee did you had a chance to test this?

chulkilee:
I'm using this branch and it is working. I think it's ready to be merged.

Using capistrano3-unicorn would be separate issue.

arthurnn:
thanks @chulkilee

PR #795, opened 2015-01-07
Remove puma and Procfile.dev

chulkilee:
It would be better to use same app server between dev and prod.

Is there any reason to use puma on dev?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -69,8 +69,6 @@ group :development do
   gem 'binding_of_caller'
   gem 'meta_request'
   gem 'foreman', :require => false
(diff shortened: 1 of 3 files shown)


TeatroIO:
I've prepared a stage. <a href='http://chulkilee-remove-puma.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com' target='_blank'>Click to open.</a>

arthurnn:
I agree on having only 1 webserver !

stevecrozz:
Sounds good to me

PR #797, opened 2015-01-08
Fix GtalkService Keeaplive-Thread leak

felixbuenemann:
The GoogleTalk/Jabber Notification Service was leaking keepalive threads, because xmpp4r creates an internal keealive thread on connect that is only cleaned up when closing the client.

This lead to workers quickly running out of threads and throwing exceptions like `ThreadError: can't alloc thread`. The fix ensures that the xmpp4r client is always closed after notifications have been sent.

Example Exception-Backtrace:
```
ThreadError: can't alloc thread:
xmpp4r-0.5.5/lib/xmpp4r/connection.rb:83→ new
xmpp4r-0.5.5/lib/xmpp4r/connection.rb:83→ connect
xmpp4r-0.5.5/lib/xmpp4r/client.rb:71→ connect
app/models/notification_services/gtalk_service.rb:45→ create_notification
mongoid-4.0.0/lib/mongoid/relations/proxy.rb:150→ method_missing
app/models/notice.rb:196→ services_notification
```

### DIFF
--- a/app/models/notification_services/gtalk_service.rb
+++ b/app/models/notification_services/gtalk_service.rb
@@ -53,6 +53,8 @@ class NotificationServices::GtalkService < NotificationService
     # post the issue to the xmpp room(s)
     send_to_users(client, message) unless user_id.blank?
     send_to_muc(client, message) unless room_id.blank?
+  ensure
+    client.close unless client.nil?
   end
 
   private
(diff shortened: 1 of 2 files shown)


TeatroIO:
I've prepared a stage. <a href='http://felixbuenemann-fix-gtalk-keepalive-thread-leak.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com' target='_blank'>Click to open.</a>

felixbuenemann:
I'll look into the test failures, looks like the test doubles don't expect close to be called.

felixbuenemann:
Specs fixed.

arthurnn:
thanks..ideally we will extract that code.. but this make sense..

Issue #784, opened 2015-01-05
plugin icon improvements


stevecrozz-:
Currently, plugins provide icons (images) for remote services by virtue of some rails engine/asset pipeline magic and forcing errbit the application to serve up a dynamically generated CSS file with directives that set background images. When plugins are changed, the css file that references them doesn't and that can cause stale cache issues. Moreover, this is an overcomplication.

Plugins should have a much more straightforward way of providing images to errbit for use as icons. Something like a method called 'icon' on the plugin itself that returns the image's binary data seems like a good way to go. This will have to change here and in errbit/errbit_plugin

stevecrozz:
@arthurnn I noticed you assigned this to yourself. Do you have everything you need on this?

arthurnn:
Yes, I do... just need time to finish it.

stevecrozz:
Here's my attempt at it:
https://github.com/errbit/errbit/compare/master...stevecrozz:new_icon_interface
https://github.com/errbit/errbit_plugin/compare/master...stevecrozz:new_icon_interface
https://github.com/errbit/errbit_github_plugin/compare/master...stevecrozz:new_icon_interface

arthurnn:
wanna send PRs on the plugin projects first..

stevecrozz:
Sure. Here's the upstream PRs:
https://github.com/errbit/errbit_plugin/pull/3
https://github.com/errbit/errbit_github_plugin/pull/5

arthurnn:
:cocktail: :cool:

stevecrozz:
Fixed in #859

PR #798, opened 2015-01-08
Add missing icons for github issue service

felixbuenemann:
These are referenced on the app preferences page.


TeatroIO:
I've prepared a stage. <a href='http://felixbuenemann-add-missing-github-icons.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com' target='_blank'>Click to open.</a>

stevecrozz:
We shouldn't need this because the images should exist here:
https://github.com/errbit/errbit_github_plugin/tree/master/vendor/assets/images

Can you explain a bit more about where they're missing and why they can't be found in the plugin?

felixbuenemann:
I was seeing 404 errors for those images. I know that recent versions of rails/sprockets only add vendor/javascripts and vendor/styles to the load path, which could explain why the images were not found. I'm not sure what's the best way to solve this. I've seen some gems that simply moved all assets to app/assets in the gem.

felixbuenemann:
Hmm, strange. I checked `Rails.application.config.assets.paths`and it lists the vendor/assets/images from the github plugin. I'll investigate further what's going on.

felixbuenemann:
The issue seems to be that images in vendor/assets are no longer precompiled automatically since rails 4.0. So the images would be found in development, but not in production. Probably the easiest fix is to rename vendor folder to app inside the errbit_github_plugin.

stevecrozz:
Thanks very much for figuring this out. I can see what's going on now. My thought is this isn't worth the effort considering we'll be overlapping with: https://github.com/errbit/errbit/issues/784 which is on the road map for 0.4.0. So I'm going to close this one.

felixbuenemann:
Fine with that.

Issue #553, opened 2013-08-29
Deploy Message via Airbrake Gem?


tresacton:
Can somebody please advise whether or not it is possible to set a deploy message? If it is possible, please explain how.

If you need clarification for what I am asking for, please do the following: 
1. Navigate to the index page for apps
2. Notice the message column? For me this is always empty.

I want to fill this value in my other application's deploy.rb script, with the result of "git log --oneline -1" for the app being deployed.

stevecrozz:
Added some documentation with #800

tresacton:
Thanks. This is a very late reply, but that's exactly what I ended up doing.

PR #800, opened 2015-01-12
#553 document deploy hooks

stevecrozz:


TeatroIO:
I've prepared a stage. <a href='http://stevecrozz-553-document-deploy-hooks.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com' target='_blank'>Click to open.</a>

shingara:
Good. But you can say it's become with capistrano too.

stevecrozz:
@shingara Do we have some special capistrano support?

stevecrozz:
Ah, you must be talking about this:
https://github.com/airbrake/airbrake/blob/master/lib/airbrake/tasks/airbrake.cap

I'll add some docs on how to use the capistrano task.

PR #804, opened 2015-01-13
Fix fingerprinting migration as it is refactored

pinglamb:
Commit https://github.com/errbit/errbit/commit/da3946bef7921e5a6df0d55fc2f556b641884ada refactored fingerprinting into a Fingerprint module, updated the migration for that.


TeatroIO:
I've prepared a stage. <a href='http://pinglamb-master.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com' target='_blank'>Click to open.</a>

PR #805, opened 2015-01-13
add exception occurance trend bars (time distribution of exceptions)

manygrams:
This PR adds trend bars to Errbit, which gives you an idea of the time distribution of an exception. These trend bars are amazingly useful for determining whether those 400 occurrences of an exception are going steady or were a one time burst.

Other than my change to `errbit.css.erb`, this is entirely the work of my colleague @jstorimer, so all credit goes to him. I found this addition really useful and wanted it in my own side project, so here we are.

It looks like this:
image
image


TeatroIO:
I've prepared a stage. <a href='http://manygrams-trendlines.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com' target='_blank'>Click to open.</a>

stevecrozz:
This is very cool, thanks! It looks like a nice feature. I do have a bit of feedback for you and I'd really appreciate the effort if you can get this in:

- What is this Chart.js thing and can we include the unminified source code instead of minified code? It's nice to have for reference and the asset pipeline can minify it just like our other javascript
- I'm a little worried that this could kill performance when you have a ton of errors. Can you add a toggle via environment variable? something like ERROR_TREND_CHART_ENABLE=true? Might need a bit more work to make sure it looks good in both enabled and disabled states
- I would prefer to not generate dynamic JavaScript. Can you put the dynamic data into data attributes (showLabels, scaleSteps) on the canvas tags and then have static JavaScript come along later and draw the charts?
- The extra JS should go into app/assets so the asset pipeline can minify and serve it all up in a (hopefully cached) javascript blob

One extra thing that could be nice:
- I'm sure the first question we'll get about this is if users can configure the charts, so maybe add some environment variables for that too like ERROR_TREND_CHART_DAYS=10

I can help out or discuss further if needed.

manygrams:
:+1: I like the suggestions! I think I can handle adding these.

tobi:
:+1:

arthurnn:
@manygrams: I will review and merge that this weekend. thanks for the PR

stevecrozz:
@arthurnn I don't think this is ready to go yet. We're waiting on some changes still. Do you think those changes are unnecessary?

app/views/shared/_bar_chart.html.erb
```diff
@@ -0,0 +1,27 @@
+<%= content_tag :canvas, nil, id: "chart-for-#{problem.id}", width: width, height: height, data: { labels: bar_chart_x_axis_labels(show_labels).to_json, notice_count: data, show_labels: show_labels, scale_steps: bar_chart_scale_steps(data.max) } %>
+
+<%= javascript_tag do %>
+  var canvas = $("#chart-for-<%= problem.id %>")
```
manygrams: I couldn't think of a good way to not generate this JS - any ideas?

manygrams:
@arthurnn @stevecrozz I think I have addressed your comments, mind taking another look?

manygrams:
Also, should I add the configuration information to the readme?

manygrams:
Whoops, I broke the tests... I'll fix that now

app/views/shared/_bar_chart.html.erb
stevecrozz: Absolutely. First, you'd need to somehow identify this canvas tag as a bar chart. That could be as simple as adding a class as in: ```<canvas class="bar-chart" data-problem-id="#{problem.id}"...>```

I'll just rough out a snippet that we could put in application.js or some similar place. This probably has errors/bugs because I'm just freehanding it, but this would make the js entirely static. The only dynamic parts are the data attributes of the canvas tag:
```javascript
$('canvas.bar-chart').each(function(){
  var canvas = $(this);
  var problemId = canvas.data('problemId');
  var ctx = canvas.get(0).getContext("2d");

  chart = new Chart(ctx);
  chart.Bar(data, {
    scaleShowLabels: canvas.data('show-labels'),
    scaleOverride: true,
    scaleSteps: canvas.data('scale-steps'),
    scaleStepWidth: 1,
    scaleStartValue: 0,
    animation: false,
    scaleShowGridLines: false,
    barValueSpacing: 1
  });
});
```

stevecrozz:
As far as docs go, it could be as simple as adding two little blurbs in the configuration section on how the new ENV variables can be used: https://github.com/errbit/errbit/blob/master/docs/configuration.md

I don't think we need more than that for this PR, but if you want to write more you could make a new page in the docs folder.

manygrams:
@stevecrozz I took another stab but I can't quite get it to work. The charts work fine in the errs table, but the charts don't show up on the show pages. It seems like the `chart.Bar` part sets the width/height to 0px. My JavaScript skills aren't very strong - would you be able to take a look and see if you see any obvious problems / have any suggestions of what I'm doing wrong?

stevecrozz:
Sure. I'll take a look

manygrams:
@stevecrozz FYI I rebased in order to edit the `configuration.md` - it must have been created after I branched

stevecrozz:
I put a PR into your PR: https://github.com/manygrams/errbit/pull/1

manygrams:
Hrmm I don't see how the test failure is related to the change - maybe a timing issue. I'll look into it

danielbeardsley:
We had the same idea!  I did this in a way that's seems lighter-weight #864 

I like that you've added them to the main problems page... how does it affect page load time?

Issue #806, opened 2015-01-14
Problem with assets on heroku


gabrielhilal:
I have followed the [docs] (https://github.com/errbit/errbit/blob/master/docs/deployment/heroku.md) to deploy `errbit` on `heroku`, however for some reason it's not loading the assets (404):
The mongo database is working fine, as I can create/remove users. Also, I can see that assets have been successfully precompiled and available in the public folder.
Please note the heroku logs:

>2015-01-14T12:48:45.806414+00:00 heroku[router]: at=info method=GET path="/" host=errbit-real.herokuapp.com request_id=351c28bf-41c1-4308-b148-f1d7f1a21a53 fwd="179.176.185.246" dyno=web.1 connect=1ms service=26ms status=304 bytes=373
2015-01-14T12:48:46.077817+00:00 heroku[router]: at=info method=GET path="/assets/application-537d72cd66bb2da9b88e7f39029a4ef4.css" host=errbit-real.herokuapp.com request_id=3eeaf8f3-cb5f-44f5-938d-f6496670df66 fwd="179.176.185.246" dyno=web.1 connect=2ms service=80ms status=404 bytes=5573
2015-01-14T12:48:46.406997+00:00 heroku[router]: at=info method=GET path="/assets/application-79e3a5026113f4ba59af151c14f3fe95.js" host=errbit-real.herokuapp.com request_id=9b766d9b-2500-41a4-8968-aa889e76d7a3 fwd="179.176.185.246" dyno=web.1 connect=4ms service=250ms status=404 bytes=5572
2015-01-14T12:48:46.779301+00:00 heroku[router]: at=info method=GET path="/" host=errbit-real.herokuapp.com request_id=e12d1e8c-1d32-49c4-84c9-75812c5fda3e fwd="179.176.185.246" dyno=web.1 connect=1ms service=145ms status=304 bytes=373
2015-01-14T12:48:47.043797+00:00 heroku[router]: at=info method=GET path="/assets/application-537d72cd66bb2da9b88e7f39029a4ef4.css" host=errbit-real.herokuapp.com request_id=7289045b-7a59-4658-b22c-fb66ace4a2fb fwd="179.176.185.246" dyno=web.1 connect=1ms service=72ms status=404 bytes=5573
[...]

PR #807, opened 2015-01-14
fix assets on heroku

gabrielhilal:
add instructions on how to include the rails_12factor gem to solve the assets issue on heroku

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -17,6 +17,7 @@
 - [@felixbuenemann][]
 - [@soberstadt][]
 - [@rsutphin][]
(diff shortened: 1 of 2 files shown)


TeatroIO:
I've prepared a stage. <a href='http://gabrielhilal-fix-heroku-assets.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com' target='_blank'>Click to open.</a>

stevecrozz:
Thanks for bringing this up. I'm going to put #806 on the 0.4.0 road map because I think we need to get it fixed, but I added some comments there because I think this is not the fix we want.

PR #808, opened 2015-01-18
Implement JSON Api v3

ronin:
This will allow to capture errors sent by airbrake-js.


TeatroIO:
I've prepared a stage to preview changes. <a href='http://ronin-json-api-v3.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com' target='_blank'>Open stage</a> or <a href='http://ronin-json-api-v3.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com/teatro.log' target='_blank'>view logs</a>.

shivanibhanwal:
Thank you so much @ronin! It is working fine with the airbrake-js . Just tested with my app and properly logged error in the errbit!

ronin:
So what do you think about this PR? I know there is room for improvement and discussion, but on the other hand it will be great to get some feedback from people running this in production.

stevecrozz:
Sorry for the delay. I think this is a great start. I'd love to see a controller test, and I'd rename JsonParser to something a little more specific. Maybe something like AirbrakeApi::V3::NoticeParser. Have you tested this with anything besides the airbrake-js client yet? I'm guessing there's some other clients we may want to test it with.

Assuming that looks good, then I think that's all we need to make this mergable. We'll need some docs too, but we can work on those later.

ronin:
Thanks for the feedback. I will do these changes shortly. As for other clients, I've only tested airbrake-js. Do you have any specific clients in mind? I think I can monkey-patch airbrake to use json api and see what happen.

stevecrozz:
@ronin. Yeah I guess you'd have to monkey patch it. Looks like the json api is only enabled with the hostname matches collect.airbrake.io:
https://github.com/airbrake/airbrake/blob/master/lib/airbrake/sender.rb#L176-L177

ronin:
Ok, I've made those changes. I've added controller tests and updated readme with information how to set airbrake-js. I did some changes so now this can work with ruby airbrake client. I've tested with rake airbrake:test for both v2 and v3 api and looks the same and is grouped correct. Only difference is that in v3 api framework attribute is missing, airbrake doesn't send it.

I've put info on how to setup airbrake to use v3 api.

stevecrozz:
Really nice work @ronin

steveyken:
great work. thank you all

shivanibhanwal:
:+1: @ronin @stevecrozz

PR #809, opened 2015-01-18
Adds the rails_12factor gem for Heroku

matiaskorhonen:
This PR adds the rails_12factor gem to the Gemfile, but it is only required if the `HEROKU` environment variable is set. This should have no ill effect on other deployment targets. This should solve issue #806…

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -93,6 +93,7 @@ group :test do
 end
 
 group :heroku, :production do
+  gem 'rails_12factor', require: !!ENV["HEROKU"]
   gem 'unicorn', :require => false
 end
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -264,8 +264,13 @@ GEM
       bundler (>= 1.3.0, < 2.0)
       railties (= 4.1.8)
       sprockets-rails (~> 2.0)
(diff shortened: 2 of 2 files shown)


TeatroIO:
I've prepared a stage to preview changes. <a href='http://matiaskorhonen-optional-rails-12factor.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com' target='_blank'>Open stage</a> or <a href='http://matiaskorhonen-optional-rails-12factor.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com/teatro.log' target='_blank'>view logs</a>.

arthurnn:
we tried to add this, and had to revert. @stevecrozz do you remember why?
BTW, the rails_12factor gem add about 3 configs only, so we might be able to do that by hand too.

matiaskorhonen:
As far as I can tell the last time was PR #765?

In that pull request the Heroku gem wasn't being conditionally loaded as it is in this case. See the last (unanswered) comment by @schneems in that PR…

schneems:
This looks good to me :+1:

arthurnn:
thats right.. thanks

PR #810, opened 2015-01-20
configuration through env only

stevecrozz:
This PR is the result of some experimentation I've been doing with Errbit's configuration system. This is a pretty drastic change and I haven't docs and tests yet, but I wanted to get some eyes on it to see if it has any value.

This removes almost all the YAML config and replaces them with a system of configuration through environment variables. All the rules for how to configure Errbit from environment variables are in config/initializers/_load_config.rb. Actual configuration no longer lives in files, except for default configuration which lives in .env.default. The idea I had in mind was that Errbit should have only one way to configure it that works for all kinds of deploys including capistrano, heroku and hopefully many others.

This also removes the cloud66-specific configuration. I don't know much about this one, so definitely chime in if there's a good reason we need to keep it. I'd just prefer a more cloud-agnostic approach to deployment, if that's possible.


TeatroIO:
I've prepared a stage to preview changes. <a href='http://stevecrozz-env-config.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com' target='_blank'>Open stage</a> or <a href='http://stevecrozz-env-config.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com/teatro.log' target='_blank'>view logs</a>.

config/initializers/_load_config.rb
```diff
@@ -96,6 +91,6 @@
   default.reverse_merge!(options_from_config)
 end
 
-if Rails.env.production?
-  Rails.application.config.consider_all_requests_local = Errbit::Config.display_internal_errors
-end
+Rails.application.config.consider_all_requests_local = Errbit::Config.display_internal_errors
```
arthurnn: I think this 3 rails config should not be part of the errbit config.
They are rails standards, and they are mostly hard-coded on the right environment file.
Just `secret_key_base` under production is one that uses a ENV var, but I would say we should still use it like that.

arthurnn:
I like where this is going. We should have only 1 way to config the app, and I would say ENV vars are a easy and flexible way.
A few comments:

- maybe we should to the config loading(`Errbit::Config = Configurator.run`) on a `before_initialize` event, http://guides.rubyonrails.org/configuring.html#initialization-events , so we make sure config are run before other libraries.
- I dont get it why we need a mongo initializer, could we just make `mongoid.yml` file read from `Errbit::Config` ?

config/initializers/_load_config.rb
stevecrozz: This idea grew out of a strict interpretation of of 12factor.net:
> Another aspect of config management is grouping. Sometimes apps batch config into named groups (often called “environments”) named after specific deploys, such as the development, test, and production environments in Rails. This method does not scale cleanly: as more deploys of the app are created, new environment names are necessary, such as staging or qa. As the project grows further, developers may add their own special environments like joes-staging, resulting in a combinatorial explosion of config which makes managing deploys of the app very brittle.

I was actually toying with the idea of removing all those named environment files also because they're basically meaningless. If everything is configured through environment variables, there's not much need for grouping config. We may as well just have one environment.rb file that includes all the truly hard coded stuff and reference Errbit::Config for everything else.

I might just do this to see what it looks like in this PR. I can always take it out if it sucks.

stevecrozz:
The mongo thing is interesting. We could put it in a yaml file, process it with ERB and let mongoid read and interpret the contents, but it seemed simpler to skip the yaml and erb steps and just configure mongoid directly. I won't push too hard on this though.

I'll also read up on the before_initialize event. That seems like good idea.

config/initializers/_load_config.rb
stevecrozz: After playing with this idea, I've convinced myself that these named environment files are not meaningless. I thought I would find a lot more similarities between between named environments.

Here's what I'm thinking now. We should allow people to configure via environment variables anything that might change between deploys within the same named environment. For example this would include serve_static_assets, log_level, log_location and secret_key_base.

I know rails likes to do the erb -> yaml dance for secret_key_base in a special file called secrets. I don't understand why that's a good idea.

arthurnn:
I like the mongoid initializer +1, but what we should do, is make sure that initializer run as soon as possible, we could do something like this:
```
    initializer "errbit.mongoid", before: "mongoid.load-config" do
# Mongoid load logic
    end
```
https://github.com/mongoid/mongoid/blob/master/lib/mongoid/railtie.rb#L58

Thats so, if there is any Model that do something in a initializer, that would work, because mongoid connects right after it doesnt find the file.

stevecrozz:
How are you feeling about this @arthurnn? I have on my list still to update the deployment docs, add to the changelog and add some notes to the 'upgrade' section for people upgrading to this change. One of the biggest changes here is that db config now must be provided in URL format rather than each separate component. I'm comfortable with that. Are you?

config/application.rb
```diff
@@ -18,6 +18,18 @@ class Application < Rails::Application
...
 
+    config.before_initialize do
+      # Load up Errbit::Config with values from the environment
+      require Rails.root.join('config/load')
+
+      config.secret_key_base = Errbit::Config.secret_key_base
+      config.serve_static_assets = Errbit::Config.serve_static_assets
+    end
+
+    initializer "errbit.mongoid", before: "mongoid.load-config" do
```
arthurnn: why `before` and not `after` ?

arthurnn:
This is good @stevecrozz . What I would do is, squash the commits in one, as there is a lot of back-and-forth on them. And you do the next tasks in other PRs. :shipit:

config/application.rb
stevecrozz: Not sure. Do you think it should be after? I'll defer to your judgement on this since you seem to be well versed in mongoid.

arthurnn: So if we have this as `after` if there is a mongoid.yml file, this configuration will override that one, which I think thats what we want, correct?

stevecrozz: Now that you mention it. I suppose we might want to allow users to provide their own yaml file if they must. It would add some flexibility to allow it. Although I would like users to raise issues if they find something about errbit/mongoid that can't be configured through environment variables.

I have mixed thoughts on this, but my inclination is to err on the side of flexibility if it doesn't hurt anything, and allow the yaml file to override this config.

stevecrozz:
Here's a squashed version of this PR. Closing this one:
https://github.com/errbit/errbit/pull/820

PR #813, opened 2015-01-22
Return false in the overridden onerror function.

tomciopp:
onerror should return false in order to allow the default browser behavior for javascript exceptions. Reference document here: https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers.onerror

### DIFF
--- a/public/javascripts/notifier.js
+++ b/public/javascripts/notifier.js
@@ -491,7 +491,7 @@ printStackTrace.implementation.prototype = {
 };// Airbrake JavaScript Notifier
 (function() {
     "use strict";
-    
+
     var NOTICE_XML = '<?xml version="1.0" encoding="UTF-8"?>' +
         '<notice version="2.0">' +
             '<api-key>{key}</api-key>' +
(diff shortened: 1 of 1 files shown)


TeatroIO:
I've prepared a stage to preview changes. <a href='http://tomciopp-js-onerror-false.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com' target='_blank'>Open stage</a> or <a href='http://tomciopp-js-onerror-false.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com/teatro.log' target='_blank'>view logs</a>.

stevecrozz:
Thanks @tomciopp .

Although, we're actually phasing out support for notifier.js. Since errbit is now compatible with airbrake api v3, we are able to use airbrake-js to catch js exceptions. I'm adding a ticket now to remove notifier.js and refer our users to airbrake-js instead: #814

This is a new development, though, and I'd appreciate any feedback you have after trying airbrake-js with errbit.

PR #822, opened 2015-01-29
add docker support

stevecrozz:
Now that we have env-only config, docker support is pretty straightforward. Is this something we should have in errbit? I also created a Docker errbit organization so we can use that instead of 'stevecrozz' if we decide to go ahead with this.

One thing left to consider is the the db create / rake errbit:bootstrap stuff.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,14 @@
-FROM ubuntu:utopic
+FROM ubuntu
 RUN apt-get update
 ADD . /errbit
 WORKDIR /errbit
 
 RUN DEBIAN_FRONTEND=noninteractive apt-get -y install \
-  ca-certificates gcc git make node ruby-dev ruby2.1 && \
+  ca-certificates gcc git make node ruby-dev ruby && \
   cd /errbit && gem install bundler && bundle install && \
   DEBIAN_FRONTEND=noninteractive apt-get -y purge gcc git make ruby-dev && \
   DEBIAN_FRONTEND=noninteractive apt-get -y autoremove
+
+RUN env > /myenv
 
 CMD ["bundle", "exec", "foreman", "start"]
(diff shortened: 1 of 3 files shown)


stevecrozz:
I think I figured out a decent way to run the bootstrap task. Since it no longer does any filesystem work, we can just run it in another container in a separate foreground process that exists and deletes itself when finished.

simonoff:
Any progress?

stevecrozz:
@imonoff I'd say the image is ready for some testing. Would you like to give it a try? I've published it to my own dockerhub account for now:
https://registry.hub.docker.com/u/stevecrozz/errbit/

simonoff:
@stevecrozz  I already tests it but don't like :) And I found one bug related to MONGO_URL env, so please look on it. And I mostly finished with proper docker image for errbit based on passenger image with cron support.

stevecrozz:
Very nice. Let me know if you have any more specific feedback about what you'd like in a docker image. Maybe you can point me to the one you made so I can have a look at. It sounds like you want an image with more features built-in like cron. I'm also interested in why you want passenger. Is unicorn insufficient for your purposes?

simonoff:
@stevecrozz for me there is no differences between unicorn and passenger. Main point is what passenger image already has all staff what i need. You can look on Dockerfile there - https://github.com/simonoff/errbit-docker

lucianosousa:
@stevecrozz I'm really interested in Docker these days. Can I help here with something or this is alrealy done?

stevecrozz:
@lucianosousa I created an image a while back (https://hub.docker.com/r/stevecrozz/errbit/). I've just been waiting to validate interest in this idea and whether we should add it to Errbit. Do you want to give it a shot and send me your feedback? The image is a bit old right now, but I don't think that will be a problem for evaluating the image.

lucianosousa:
Yes, sure. I'll.

This weekend I hope test it.

stevecrozz:
Superceded by #1108

PR #826, opened 2015-02-09
Update seed: random password during install

rud:
This seems superior to the default password of `'password'`. Thoughts on this?

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,17 +1,19 @@
+require 'securerandom'
+
 puts "Seeding database"
 puts "-------------------------------"
 
 # Create an initial Admin User
 admin_username = "errbit"
 admin_email = "errbit@#{Errbit::Config.host}"
-admin_pass  = 'password'
+admin_pass  = SecureRandom.urlsafe_base64(12)[0,12]
 
 puts "Creating an initial admin user:"
 puts "-- username: #{admin_username}" if Errbit::Config.user_has_username
 puts "-- email:    #{admin_email}"
 puts "-- password: #{admin_pass}"
 puts ""
-puts "Be sure to change these credentials ASAP!"
+puts "Be sure to note down these credentials now!"
 user = User.find_or_initialize_by(:email => admin_email) do |u|
   u.name = 'Errbit Admin'
   u.password = admin_pass


TeatroIO:
I've prepared a stage to preview changes. <a href='http://rud-patch-1.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com' target='_blank'>Open stage</a> or <a href='http://rud-patch-1.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com/teatro.log' target='_blank'>view logs</a>.

db/seeds.rb
```diff
@@ -1,17 +1,19 @@
+require 'securerandom'
+
 puts "Seeding database"
 puts "-------------------------------"
 
 # Create an initial Admin User
 admin_username = "errbit"
 admin_email = "errbit@#{Errbit::Config.host}"
-admin_pass  = 'password'
+admin_pass  = SecureRandom.hex
```
arthurnn: maybe something smaller `SecureRandom.hex[0..7]` ? 8 digts is enough i would say..

rud:
Certainly possible - but since it's a value you have to copy/paste somewhere anyway (noone is going to memorize it), the length does not truly matter, and longer is always better with passwords. 

By default it's a 32-character strongly random value. That can safely be halved if that seems more manageable:
```ruby
>> SecureRandom.hex
=> "290bc5e2dccfa22488310a42f2582dc7"
>> SecureRandom.hex[0...16]
=> "5799df7f0b2eaf49"
```

Thoughts?

rud:
There's more bits per character with `SecureRandom.urlsafe_base64`:
```ruby
>> SecureRandom.urlsafe_base64(12)[0,12]
=> "5GrwM5U9Fl7D"
>> SecureRandom.urlsafe_base64(12)[0,12]
=> "pWSTm8sZiQJD"
>> SecureRandom.urlsafe_base64(12)[0,12]
=> "YGDec29adJmQ"
```

That would allow the value to be safely shorter while keeping it nice and complex.

arthurnn:
yeah, i like `SecureRandom.urlsafe_base64(12)[0,12]`

rud:
@arthurnn cool, I've pushed an update to this pull-request, hope it is useful!

arthurnn:
thanks

rud:
Awesome, thank you!

Issue #785, opened 2015-01-05
better default issue body template


stevecrozz-:
The current template at app/views/issue_trackers/issue.txt.erb is not a great default template because it contains a mix of mardown and html which is not a very sane default for an issue tracker body. This should probably just be a simple, plain-text template instead, at least to start with. We can add others later.

arthurnn:
:+1: ,

In the future we potentially could have multiple types of templates, things like
app/views/issue_trackers/issue.txt.erb
app/views/issue_trackers/issue.html.erb
app/views/issue_trackers/issue.md.erb

And the issue tracker could leverage the rails render format to render the right issue body.

stevecrozz:
I think that's a nice idea. I think its more than what we need for 0.4.0, but I'd be happy to have these.

stevecrozz:
@arthurnn I added a PR #785 which has a txt and md issue tracker template. I made the default md since that's what we'll use for github.

This is on the 0.4.0 milestone: https://github.com/errbit/errbit/milestones/0.4.0

stevecrozz:
Fixed with #783

Issue #783, opened 2015-01-05
allow issue tracker gems to provide a template for the remote issue body


stevecrozz-:
As of #774, issue trackers can only have one template and that is not necessarily suitable for all issue trackers. Errbit should allow issue tracker plugins to provide their own template.

stevecrozz:
This one is ready to start. Now that #785 is done, it may make sense to add the ability for issue trackers to specify a default template to use (md, txt, etc). I think we still need a way for plugins to define the entire template, but the default template should work for most trackers since we now have multiple formats.

Here's one idea for an interface that should work:
IssueTracker#body_template **optional** if this method exists, it should return an ERB template to render, or a symbol. If it is a symbol it should be one of the available default template formats such as txt or md (like the ones I created in #774). If this method is unimplemented, we'll render the default template in markdown format.

stevecrozz:
@arthurnn Added PR #840 which should give us what we want here

PR #827, opened 2015-02-09
Refs #785 add md and txt issue tracker templates

stevecrozz:
This PR adds an unused txt format issue tracker body and a markdown format issue tracker body to prepare for issue trackers that can specify which template to use.

### DIFF
--- a/app/controllers/problems_controller.rb
+++ b/app/controllers/problems_controller.rb
@@ -61,7 +61,7 @@ class ProblemsController < ApplicationController
   end
 
   def create_issue
-    body = render_to_string "issue_trackers/issue", layout: false, formats: [:txt]
+    body = render_to_string "issue_trackers/issue", layout: false, formats: [:md]
     title = "[#{ problem.environment }][#{ problem.where }] #{problem.message.to_s.truncate(100)}"
 
     issue = Issue.new(problem: problem, user: current_user, title: title, body: body)
--- a/app/helpers/hash_helper.rb
+++ b/app/helpers/hash_helper.rb
@@ -1,6 +1,8 @@
 module HashHelper
 
   def pretty_hash(hash, nesting = 0)
+    return '{}' if hash.empty?
+
     tab_size = 2
     nesting += 1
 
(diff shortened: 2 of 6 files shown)


TeatroIO:
I've prepared a stage to preview changes. <a href='http://stevecrozz-issue-tracker-templates.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com' target='_blank'>Open stage</a> or <a href='http://stevecrozz-issue-tracker-templates.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com/teatro.log' target='_blank'>view logs</a>.

arthurnn:
make sense to me!

app/controllers/problems_controller.rb
```diff
@@ -61,7 +61,7 @@ def show
   end
 
   def create_issue
-    body = render_to_string "issue_trackers/issue", layout: false, formats: [:txt]
+    body = render_to_string "issue_trackers/issue", layout: false, formats: [:md]
```
arthurnn: what I want in the future is,probably, have the issue tracker doing a POST to this endpoint passing the right format, so we can render with different formats depending on what the issue trakcer supports.

stevecrozz: Can you explain that a bit more? Right now this is the action that we execute when you click 'create issue' from an issue page. Are you imagining a totally different flow?

arthurnn: No, I imagine the same, so that action will send an extra parameter like `issue_format=md` , so we can decide which format we will render the issue body..
And when we render the format, we add this parameter based on the issue_tracker engine, which will provide a `body_format` method. or something like that.

stevecrozz: Sounds good. I'm thinking we could even potentially leave that decision to Issue#save, where we have everything we need to make that happen (the issue tracker object and the problem object). But we can probably solve this as part of https://github.com/errbit/errbit/issues/783

I'll ping you from that one and we can decide what to do.

arthurnn:
I think this is good to go,
@stevecrozz ?!

stevecrozz:
Yeah let's do this. Thanks for the review

PR #830, opened 2015-02-11
Fix parsing V3 api notice without backtrace

### LINKED ISSUE #828: Error in Api v3 NoticeParser when backtrace is nil
https://github.com/errbit/errbit/blob/master/lib/airbrake_api/v3/notice_parser.rb#L35

When submitting over the v3 API (airbrake-js) Errbit raises an error if the backtrace is nil.

    undefined method `map' for nil:NilClass

An error without a backtrace isn't very helpful in the first place, so I'll be fixing the main problem in my client system that is sending an empty backtrace. However, I thought it worth raising an issue here for Errbit to gracefully handle bad input like this.

As a side note, really appreciate the work that was done on the v3 API recently. It's made a real difference for us and I know the paint is still drying.

ronin:
Fixes #828.

### DIFF
--- a/lib/airbrake_api/v3/notice_parser.rb
+++ b/lib/airbrake_api/v3/notice_parser.rb
@@ -32,7 +32,7 @@ module AirbrakeApi
       end
 
       def backtrace
-        error['backtrace'].map do |backtrace_line|
+        (error['backtrace'] || []).map do |backtrace_line|
           {
             method: backtrace_line['function'],
             file: backtrace_line['file'],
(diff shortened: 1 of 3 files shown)


TeatroIO:
I've prepared a stage to preview changes. <a href='http://ronin-fix-empty-backtrace.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com' target='_blank'>Open stage</a> or <a href='http://ronin-fix-empty-backtrace.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com/teatro.log' target='_blank'>view logs</a>.

stevecrozz:
Thanks @ronin.

I noticed there is a json schema for this blob, so we could explore validating the incoming request body against the json schema which in this case would not allow a missing backtrace, but I think this is fine.

ronin:
We could but Isn't it a bit too much?

stevecrozz:
Yeah, its just an option if we need to go that route. It could be worth considering if we find any more unhanded exceptions that could be prevented with a schema check. But it may not be necessary.

PR #831, opened 2015-02-12
github_org_id can't match (String vs. Integer)

jlecour:
The user's organization ids are converted to String, but the configuration value `Errbit::Config.github_org_id` is always an Integer. They can't possibly match.

Unifying value types make this work.

### DIFF
--- a/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/users/omniauth_callbacks_controller.rb
@@ -8,8 +8,8 @@ class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
       # See if they are a member of the organization that we have access for
       # If they are, automatically create an account
       client = Octokit::Client.new(access_token: github_token)
-      org_ids = client.organizations.map { |org| org.id.to_s }
-      if org_ids.include?(github_org_id.to_s)
+      org_ids = client.organizations.map { |org| org.id }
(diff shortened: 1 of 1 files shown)


TeatroIO:
I've prepared a stage to preview changes. <a href='http://jlecour-patch-1.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com' target='_blank'>Open stage</a> or <a href='http://jlecour-patch-1.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com/teatro.log' target='_blank'>view logs</a>.

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -8,7 +8,7 @@ def github
       # See if they are a member of the organization that we have access for
       # If they are, automatically create an account
       client = Octokit::Client.new(access_token: github_token)
-      org_ids = client.organizations.map { |org| org.id.to_s }
+      org_ids = client.organizations.map { |org| org.id }
       if org_ids.include?(github_org_id)
```
arthurnn: I would also make sure that `github_org_id` is a Integer and call `github_org_id.to_i`

jlecour:
In fact, there is a test that makes sure that numbers from the config are returned as Integer. That's why I didn't coerce here.

arthurnn:
make sense then... thanks

PR #821, opened 2015-01-29
update deployment docs with env-config instructions

stevecrozz:
Heroku changes are pretty straightforward, but I tested capistrano deployment for the first time and found it to be lacking in both functionality and documentation. For example, I found no existing way to actually start the application using capistrano, so I added a task for that.

PR #832, opened 2015-02-12
Don't link and copy old config files

jlecour:
[This change](427ab3b0a0f9a03c60bb019dc1be28793a284bc4) is really great and is a huge simplification of the deployment/configuration process.

Although, the `config/deploy.example.rb` still mention the copy and linking of `config.yml` and `mongoid.yml`

I've removed them in my `config/deploy.rb` and it works.

### DIFF
--- a/config/deploy.example.rb
+++ b/config/deploy.example.rb
@@ -20,8 +20,6 @@ set :ssh_options, forward_agent: true
 
 set :linked_files, fetch(:linked_files, []) + %w(
   .env
-  config/config.yml
-  config/mongoid.yml
   config/newrelic.yml
 )
 
@@ -42,8 +40,6 @@ namespace :errbit do
     on roles(:app) do
       execute "mkdir -p #{shared_path}/config"
       {
-        'config/config.example.yml' => 'config/config.yml',
-        'config/mongoid.example.yml' => 'config/mongoid.yml',
         'config/newrelic.example.yml' => 'config/newrelic.yml'
       }.each do |src, target|
         execute "if [ ! -f #{shared_path}/#{target} ]; then cp #{current_path}/#{src} #{shared_path}/#{target}; fi"


TeatroIO:
I've prepared a stage to preview changes. <a href='http://jlecour-patch-2.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com' target='_blank'>Open stage</a> or <a href='http://jlecour-patch-2.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com/teatro.log' target='_blank'>view logs</a>.

arthurnn:
thanks!

stevecrozz:
@jlecour I've actually got a related change over in https://github.com/errbit/errbit/pull/821.

If you're using capistrano, would you mind having a look and maybe giving it a test?

arthurnn:
:heart: @stevecrozz

jlecour:
To be honest, I think that @stevecrozz 's patch is better than mine, at least, it takes care of more things.

stevecrozz:
@jlecour I can merge mine also, but I've been a little gun-shy since I don't actually use capistrano (although I did some testing locally). I'll just have to merge it soon if I don't hear of any problems.

jlecour:
@stevecrozz Being cautious is wise ;)

I'm using Capistrano and I'll try your patch too. Could you rebase it on the current master? That would help me verify that everything works.

Thanks

jlecour:
Oops, I've just seen that you did the merge.

I'll try it today.

jlecour:
@stevecrozz I've made a [PR](https://github.com/errbit/errbit/pull/833) with a few minor tweaks on your patch.
It seems to work for me, on a real setup with Capistrano, Unicorn, …

jlecour:
I have a more advanced Unicorn recipe that I plan to release soon.

I don't know if I'll make a full gem, or just a path that I'll submit here and there.

jlecour:
So, I've made a gem : https://github.com/jlecour/capistrano-unicorn

Sadly,  there are other gems with the same name and approximately the same features.
I'll have to take a deeper look to decide what to do : maintaining my own (renamed) version, or talk with the maintainers of those other gems and try to merge all this.

Your opinion is welcome.

stevecrozz:
Interesting... I wouldn't mind letting a gem handle some more responsibility here. But it also seems that there's a line where Errbit's responsibility must end. I'm not much of an ops person, but I'd imagine a real environment would need supervisord/systemd/upstart/etc scripts and I don't see a good one-size-fits-all approach.

But if there's a nice gem out there that can relieve errbit of some responsibility while not getting in the way, I'd be happy with that. I'm also not glued to unicorn. If there's better community support out there for another app container, that's something we can explore.

jlecour:
@stevecrozz I completely agree with that

 I love Unicorn, and Capistrano, but they're not part or Errbit.

I guess even stripping out the Capistrano part would make a lot of sens, if we could provide a simple drop-in gem to handle this for the common case (Capistrano + a "blessed" app server").

A few days ago, Heroku has even replaced Unicorn with Puma as their favorite app server, and since Heroku is a good fit for Errbit, maybe there is a sign here.

And besides Heroku (which I don't use) there is also the possibility of a container appraoche (Docker or else) that also makes a lot of sens. In that situation, providing a complete solution would be nice, à la Discourse. I don't know if there is that kind of potential and desire in the Errbit community, to be able to build this. But I can see that as a possible outcome.

PR #834, opened 2015-02-14
Asynchronous problem sending notifications

Strnadj:
We face one simple problem, our monitored application are depend to other services (for example SSO services, database clusters or other services). We have quite big amount of users, so if the depending service is accidentally down our errbit server is overloaded with requests (it seems in most case puma forks are waiting to notification service IO, http requests), actually we need to know about unavailability 3rd party services (so thats why we can't blacklist Timeout or Net errors in Rails applications) and we also require accepting other errors from applications. Using sidekiq with delay queue can be a quite nice solution to speed-up api responses.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -49,6 +49,13 @@ gem 'httparty'
 # Flowdock
 gem 'flowdock'
 
+# Asynchronous JOB
+# ---------------------------------------
+gem 'sidekiq', '~> 3.2.6'
+gem 'sinatra', require: false
+gem 'slim'
+gem 'sidekiq-failures'
+
 # Authentication
 # ---------------------------------------
 # GitHub OAuth
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -86,6 +86,8 @@ GEM
       rack (>= 1.0.0)
       rack-test (>= 0.5.4)
       xpath (~> 2.0)
+    celluloid (0.15.2)
+      timers (~> 1.1.0)
     cliver (0.3.2)
     coderay (1.0.9)
     coffee-rails (4.1.0)
@@ -262,6 +264,8 @@ GEM
     rack (1.5.2)
     rack-contrib (1.1.0)
       rack (>= 0.9.1)
+    rack-protection (1.5.3)
+      rack
     rack-ssl (1.4.0)
       rack
     rack-ssl-enforcer (0.2.6)
@@ -293,6 +297,9 @@ GEM
     rake (10.4.2)
     rdoc (4.1.2)
       json (~> 1.4)
+    redis (3.2.0)
+    redis-namespace (1.5.1)
+      redis (~> 3.0, >= 3.0.4)
     request_store (1.1.0)
     responders (1.1.2)
       railties (>= 3.2, < 4.2)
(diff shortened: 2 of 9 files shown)


TeatroIO:
I've prepared a stage to preview changes. <a href='http://netbrick-async-sending.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com' target='_blank'>Open stage</a> or <a href='http://netbrick-async-sending.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com/teatro.log' target='_blank'>view logs</a>.

arthurnn:
This is a good idea, however we cannot merge this right now. We are trying to remove the notification system from inside errbit core and extract that to a plugin, so people can implement this like this outside errbit.
Thanks a lot for the PR, I will keep it open, but until we finish that task, we cannot merge it.

thanks again.

Strnadj:
@arthurnn Actually in current state you are able, to move, I think almost every notification *plugins* into separate modules / gems, but not the basic NotificationService provider class... I figure it out several weeks ago, how to extract notification service provider to a separate gem, maybe I can show you how or write a simple example, extraction all services into separate gems is a matter of few hours I suppose :+1:   ...

stevecrozz:
So far we've been able to get along without any background workers which has kept us from doing anything like this. I'd like to explore alternative solutions before we commit to adding a hard dependency on redis and background workers.

I think we should take this to the mailing list and discuss the options. Would you care to start a thread there?

stevecrozz:
I've been thinking about this some more, and I think we should explore something a bit more lightweight. I do really like the idea of async notifications, but I'm not convinced that we need a full background worker process for delivering them. We should be able to do async notifications using plain old event machine, celluloid, or we could use a nicer layer on top like https://github.com/brandonhilkert/sucker_punch

Ideally, we could have async notifications with no added deployment complexity. As long as we don't need to make delivery guarantees and we just do this as a 'best-effort' type thing (which is no worse than what we have now), then something along these lines should suffice.

Strnadj:
@stevecrozz Still Iam thinking the best solution is to upgrade Rails to 4.2 and use *native-rails-background* jobs, but Iam not sure if mongoid support 4.2 yet..

arthurnn:
Latest Mongoid works with rails 4.2 . I can try to updade errbit to use those versions.
I think it is good idea to use latest ActiveJob to do this.
I am still not sure if this should be in here or in the notification extraction gem(which was not done yet)

stevecrozz:
I need to do some reading then so I can get up to speed with this new rails feature. It looks like we still are going to need a way to process these jobs though (the ActiveJob adapter), and it would be awesome if whatever we end up with could live within the current deployment scenario. That is, a single mongo (or mongo cluster), and N homogeneous rainbows processes.

I'll bend on this if it ends up being a huge pain, but I think there's enough flexibility here to get us to async greatness.

stevecrozz:
@Strnadj errbit/master is on Rails 4.2 and Mongoid 5. Are you still interested in contributing to this effort? We could talk it over in gitter... https://gitter.im/errbit/errbit

Strnadj:
@stevecrozz Yes, why not :)

stevecrozz:
@Strnadj is going to put together a new PR using activejob

PR #839, opened 2015-02-19
adding readme changes for local install

mtmcfarl:
I followed the steps word for word and got errors because of mongodb not installed / running.  I just made a few simple notes in the README for the local install and updated to the shortcut version of starting a rails server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,10 @@ gem install bundler
 bundle install
 ```
 
-  * Bootstrap Errbit. This will seed the database.
+  * Install MongoDB. Follow the directions [here](http://www.mongodb.org/display/DOCS/Ubuntu+and+Debian+packages), then:
+
+  * Bootstrap Errbit. This will seed the database.  Make sure you copy the
+    username and password down someplace safe.
 
 ```bash
 rake errbit:bootstrap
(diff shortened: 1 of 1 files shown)


TeatroIO:
I've prepared a stage to preview changes. <a href='http://codemancode-readme-changes.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com' target='_blank'>Open stage</a> or <a href='http://codemancode-readme-changes.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com/teatro.log' target='_blank'>view logs</a>.

stevecrozz:
Looks like the instructions to install mongo are already here, but above in the 'Installation' section. Was it unclear that those instructions apply to local installation too?

mtmcfarl:
Yeah, because I skipped down to the local install section and there wasn’t any mention of installing mongo.  I think its best even if it is redundant.  I think it makes it more clear, but ultimately its up to y’all.

Thanks!

stevecrozz:
I'll go ahead and merge the change. I need to do some major docs maintenance here pretty soon, so hopefully I can make this all much more clear when I get to it.

Issue #756, opened 2014-11-12
POSSIBLE BUG: Slack integration not working


henry74:
I think Slack may have changed the url for incoming webhooks.  I created a incoming webhook on slack and the webhook URL is: https://hooks.slack.com/services/T05Z3C4KT/B02V6Z0NN/randomcharactershere

If you click on the Slack icon on the Apps page it takes you to:
https://yoursubdomain.slack.com/services/hooks/incoming-webhook?token=token123here

Thoughts?

micred:
It doesn't work, even setting the token as the last characters of the webhook.
Any ideas?

PR #842, opened 2015-02-25
Fix Slack Integration Service

jules2689:
The issue
---
 - It seems that Slack stopped using the url format previously specified `https://#{subdomain}.slack.com/services/hooks/incoming-webhook?token=#{api_token}`

The fix
---
 - Instead what is now being used is the Slack Integration for Airbrake.
   - This provides the end user with a hook url (service url) in the following format `https://hooks.slack.com/services/XXXX/XXXX/XXXX`
   - They encourage this to be copied directly into Airbrake, so I figured this would be better to do with Errbit as well.

I've tested this on my local Errbit instance with great success :)

cc @arthurnn for review?


TeatroIO:
I've prepared a stage to preview changes. <a href='http://jules2689-master.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com' target='_blank'>Open stage</a> or <a href='http://jules2689-master.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com/teatro.log' target='_blank'>view logs</a>.

jules2689:
No idea why this is conflicting, master is up to date.

stevecrozz:
Without this change, slack notifications are broken?

jules2689:
From what I could tell, yes. I could not find an option in Slack to provide me with an API Token such that the old slack service could function. I tried: Incoming Webhook & Airbrake, both provided a url similar to the one in the description.

The moment I switched to this however, everything worked.

jules2689:
Related: https://github.com/errbit/errbit/issues/756
They seem to indicate the last numbers as the "API Token", however myself and another in that thread could not get that to work.

stevecrozz:
We should make this change then. I've had a small concern that this would break for anyone who has these notifications all set up and working the old way, but there's not much we can easily do for them. Thanks for this, and thanks for the added docs! Will merge momentarily.

jules2689:
I considered that as well, but I'm wondering if its even working for them at this point in time.
Bonus is that this is a lot easier to setup, and requires only one field.

PR #845, opened 2015-03-09
remove foreman

chulkilee:
[foreman](https://github.com/ddollar/foreman) says it should not be in Gemfile

> Ruby users should take care not to install foreman in their project's Gemfile.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -75,7 +75,6 @@ group :development do
   gem 'better_errors'
   gem 'binding_of_caller', platform: 'ruby'
   gem 'meta_request'
-  gem 'foreman', :require => false
 end
 
 group :test do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -136,9 +136,6 @@ GEM
     flowdock (0.3.1)
       httparty (~> 0.7)
       multi_json
-    foreman (0.77.0)
-      dotenv (~> 1.0.2)
-      thor (~> 0.19.1)
     haml (4.0.3)
       tilt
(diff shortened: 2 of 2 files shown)


TeatroIO:
I've prepared a stage to preview changes. <a href='http://chulkilee-remove-foreman.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com' target='_blank'>Open stage</a> or <a href='http://chulkilee-remove-foreman.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com/teatro.log' target='_blank'>view logs</a>.

stevecrozz:
Is there a specific problem you're having with foreman in the Gemfile? I rather like being able to lock down the foreman version on a per-project basis rather than having a system-wide foreman.

jlecour:
Let's say that you want to use `dotenv > 1` (version 2 has been released recently) : Bundler would prevent from doing that, because Foreman depends on `dotenv ~> 1.0.2`.

This dependency conflict is completely artificial and unnecessary : 

- executing foreman is completely separate from executing the rest of the software (Errbit in our case) : no code is shared
- having usage instructions to benefit from using Foreman should be in the Readme, not in the software dependency checks (in my opinion).

jlecour:
And as far as I can tell, Foreman is not required by Errbit to run properly. It's just a convenience.

stevecrozz:
OK. It seems like I'm in the minority and I understand the complaint so I will capitulate. Thanks for bringing this up again and helping to make Errbit better.

PR #864, opened 2015-03-27
Add sparklines to graph recent occurrences of a problem

danielbeardsley:
Now we include a sparkline of the timestamps from recent notices.
Two graphs give at-a-glance access to near and far history.

image

### DIFF
--- a/app/assets/stylesheets/errbit.css.erb
+++ b/app/assets/stylesheets/errbit.css.erb
@@ -740,6 +740,26 @@ table.tally th.value {
   text-transform: none;
 }
 
+/* Sparklines */
+.spark {
+  position: relative;
+  border: 1px solid #ddd;
+  float: left;
+  padding: 1px;
+  margin: 0;
+  height: 30px;
+}
+
+.spark > i {
+  display: inline-block;
+  position: relative;
+  width: 10px;
+  background: black;
+  margin: 0;
+  padding: 0;
+  vertical-align: bottom;
+}
+
 /* Resolve Errs */
 #action-bar .resolve:before {
   font-family: FontAwesome;
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -54,6 +54,32 @@ module ApplicationHelper
     collection.to_a[head_size..-1].to_a
   end
 
+  def sparkline(times, start_time, num_buckets = 10)
+    times = times.select {|time| time >= start_time }
+    buckets = bucketize(times, start_time, Time.now.to_i, num_buckets)
+    bars = buckets.map do |val|
+      percent = (val * 100.0).round(2).to_s + "%";
+      "<i style='height:#{percent}'></i>"
+    end.join()
+    "<div class='spark'>#{bars}</div>".html_safe
+  end
+
+  def bucketize(values, min, max, num_buckets)
+    get_min = lambda { |a,b| a > b ? b : a }
+    range = max - min
+    buckets = Array.new(num_buckets, 0)
+    values.each do |val|
+      normalized = (val - min) / range.to_f
+
+      # Use get_min() here because there will inevitably be one value == max
+      # which means normalized = 1 and bucket[1 * num_buckets] doesn't exist
+      bucket_index = get_min.call((normalized * num_buckets).floor, num_buckets-1)
+      buckets[bucket_index] += 1
+    end
+    max = buckets.max.to_f
+    return buckets.map {|count| count / max}
+  end
+
   def issue_tracker_types
     ErrbitPlugin::Registry.issue_trackers.map do |_, object|
       IssueTrackerTypeDecorator.new(object)
--- a/app/models/problem.rb
+++ b/app/models/problem.rb
@@ -207,8 +207,15 @@ class Problem
 
     # recache each new problem
     new_problems.each(&:recache)
-
     new_problems
+  end
+
+  # Returns an array of unix epoc timestamps for the created_at field of
+  # all notices sincethe since parameter
+  def timestamps_since(since)
+    notices.where(:created_at.gte => since).
+    pluck(:created_at).
+    map {|created_at| created_at.to_time.to_i }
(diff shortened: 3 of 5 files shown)


TeatroIO:
I've prepared a stage to preview changes. <a href='http://ifixit-sparklines.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com' target='_blank'>Open stage</a> or <a href='http://ifixit-sparklines.23f39035947abfbf29f9f7741cb9c271.ttrcloud.com/teatro.log' target='_blank'>view logs</a>.

rud:
This looks like a pretty awesome addition, great work. @danielbeardsley would you be willing to write a few tests for it?

danielbeardsley:
Sorry it's been so long. I've merged master in. I still don't have bandwidth for a figuring out a test. Though it's been reliable in our almost 2 years of using it :-)

stevecrozz:
I've been revisiting this every few weeks for a while and I think I at least owe the community an answer about why I haven't merged this or the other graph code PR.

The idea of displaying graphs is a good one so there's a big +1 for the concept. But I have some concerns that want to share:

First, fetching the data to drive the graphs is a potential performance issue. This could probably be alleviated by adding limits to the queries we run to populate graph data. An answer I often get for these concerns is to make the feature configurable, but I try to keep configuration down to a minimum, so we should make sure this works well and works well for all kinds of environments, even if there are zillions of notices in the DB.

Second, while the graph code proposed here looks simple and honestly very slick, I have a fear that by adding it, we could be committing to embark on a journey of building and maintaining a whole graphing library written from scratch in Ruby/HTML. Its a cool idea, but also not something this project can own. I'm afraid the problem is more complicated than it looks and we may be better off using an off-the-shelf library of some kind. Now, I'm also a big fan of things that are lightweight and simple, so it could be the case that the merits of this approach outweigh the other options.

Last, for whatever solution we wind up with, we do need some manner of testing. If we use a third-party lib, it may be enough to add some unit tests that ensure we're correctly gathering data from the db and passing it to the library in the expected format. If we have a home grown solution, then we likely need tests also for graph display.

I'm willing to be persuaded, but these are my concerns.

danielbeardsley:
Thanks for your feedback!

> First, fetching the data to drive the graphs is a potential performance issue.

True... hmm.. I thought errbit was loading *all* the notices to drive the other parts of the output, but looking again, I see that *maybe* we aren't and that the bucketed values are cached on the Problem object. We *could* cache the counts as well, but maintaining the cached counts would increase the complexity a decent amount. We could also alter the query to only select the one attribute we care about (instead of the whole Notice).

> I have a fear that by adding it, we could be committing to embark on a journey of building and maintaining a whole graphing library

I hear ya. The existing minimal implementation has been meeting our needs for several years now, so I no longer fear feature-creep on the graphing aspect as a real concern (for us).

> we do need some manner of testing.

The rest of the app is pretty well tested, so I understand. That said, I just don't have the bandwidth to dive into this, setup the fixtures and write the tests. This was a proof-of-concept solution that ended up working great and we haven't touched it in long time.

danielbeardsley:
> First, fetching the data to drive the graphs is a potential performance issue.

For what it's worth, this has never been an issue. It's only selecting the last two weeks of notices. Even when that amounts to several thousand notices, performance just hasn't been a problem.

burnettk:
Note that this UI is only on the problems/show page, not problems/index or other list views (as I originally assumed). We're test-driving this on our various errbit instances, it's awesome and feels performant (though I didn't benchmark), and I'd support getting it back into upstream with or without tests. There's also a small possibility we'd write tests for it, but the test you'd really want (does it look reasonable) is hard. It'd be easier but less useful to assert that the correct html gets onto the page. The conflict in config/locales/en.yml is very minor, as well.

rud:
Good analysis @burnettk, would you be willing to bring this changeset up to date?

burnettk:
yeah, sure, @rud ; i think the only option is to file a new PR, since I can't edit this branch and don't have access to the upstream? it is maybe noticeably slower when there are millions of occurrences. :/

danielbeardsley:
> would you be willing to bring this changeset up to date?

I'm the author here and would definitely be willing.

>  it is maybe noticeably slower when there are millions of occurrences. :/

It should only load the occurrences from the last two weeks, so that should help. Also, errbit doesn't seem to do very well (for me) when notice counts are > 100k anyway.

danielbeardsley:
Rebased on master!

rud:
Thank you for the rebase @danielbeardsley! 

Since this PR was initially made, rubocop has been added to CI, so a few additional minor code cleanups will be necessary before everything is green. Hope the errors on CI make sense, otherwise please ask

danielbeardsley:
> before: 8 sec
after: 243 sec

Woah!

Was this because all million notices were in the last 2 weeks? Or is this column not indexed (or the mongo equivalent) thus requiring a table-scan?

danielbeardsley:
I can't reproduce the long load times because all my problem pages either return quickly or never finish loading (with or without these changes).

That said, this diff (dropping the sort order) should ensure the index is used:
```
diff --git a/app/models/problem.rb b/app/models/problem.rb
index eb09636..58328bd 100644
--- a/app/models/problem.rb
+++ b/app/models/problem.rb
@@ -126,7 +126,9 @@ class Problem
   # Returns an array of unix epoc timestamps for the created_at field of
   # all notices sincethe since parameter
   def timestamps_since(since)
-    notices.where(:created_at.gte => since).
+    Notice.for_errs(errs).
+    where(:created_at.gte => since).
     pluck(:created_at).
     map {|created_at| created_at.to_time.to_i }
   end
```

burnettk:
i tried with that patch, @danielbeardsley , but it didn't affect the slow-load-time symptom (251 sec this time).

rud:
This still looks neat. Anyone willing to carry it that last bit of the way to be ready to merge?

rud:
Replaced by #1315

PR #866, opened 2015-04-09
Make EMAIL_DELIVERY_METHOD agnostic.

simonoff:
Make possibility to set EMAIL_DELIVERY_METHOD as symbol and as string. Small refactoring.

### DIFF
--- a/config/initializers/action_mailer.rb
+++ b/config/initializers/action_mailer.rb
@@ -1,5 +1,6 @@
 # Get the email delivery method and transform it to symbol
-delivery_method = Errbit::Config.email_delivery_method.to_sym
+delivery_method = nil
+delivery_method = Errbit::Config.email_delivery_method.to_sym if Errbit::Config.email_delivery_method.present?
 
 # Set SMTP settings if given.
 if delivery_method == :smtp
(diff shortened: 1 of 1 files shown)


stevecrozz:
I've been resisting stuff like this because I prefer having one correct way to do configuration, but you are the second person to raise an issue about this, so I think we should do something.

This one PR doesn't bother me, but I think we could get more value out of input validation, rather than coercion This way errbit could just die at boot time when its not properly configured and we could catch all kinds of tricky bugs. I'll see if I can come up with something agreeable.

simonoff:
@stevecrozz how you will check what it was configured properly or not? It's an additional bunch of code what is not valuable in this case. I think would be better to remove support for symbols and using only strings.

simonoff:
@stevecrozz die time is not good for docker for example.

stevecrozz:
I don't understand your arguments, but I'm swayed by the fact that more than one person has run into trouble here. So I made a commit to do the same thing you're suggesting, but in a different place:
https://github.com/errbit/errbit/commit/3b534e9c97c94063f77e8f6adfa3203bbc3f1d72

So you should be able to set EMAIL_DELIVERY_METHOD to a string now.

simonoff:
I think what main issue what you thinks in case of YAML instead of thinking in case of ENV variable. In case of YAML `:string` will be converted to `Symbol`. So you still using YAML inside of configuration for support configuration in YAML file in parallel to ENV?

Issue #868, opened 2015-04-10
Change deploys from embedded relation to has_many


ronin:
I would like to revise current errbit api for fetching data and add api for fetching deploys. The problem is deploys are embedded inside app documents, so I can't query whole deploy "table".

I can migrate this relation to has_many, but I want to know first what do you think about this change.

stevecrozz:
The idea is nice, and I think we can track deploys separately, but we need to embed into the app enough information to use in all the typical UI and API calls that involve apps. In other words, when we're listing apps, we don't want to have to separately look up a bunch of deploy records. This is mongo after all, and we should play to its strengths.

So if you want to implement a separate deploys table, that's fine with me as long as we continue tracking information about the most recent deploy in the apps table, along with any other deploy-information needed for regular site browsing activities. That might be nothing, but I'm not sure yet.

One thing to consider is if we record deploy information in two places, it could get out of sync, so we either have to be ok with that, or implement something to try to keep this stuff in sync. I'm inclined to say we should just be ok with potentially out-of-sync deploy info.

stevecrozz:
The deployment tracking feature is on the chopping block with #913, so I'm going to close this feature and we can consider deployment tracking to be deprecated at this point. Please get in touch with me if you're really attached to this feature because there's still time to talk it over.

ronin:
No, I am not attached. It is better to focus on more important features.

PR #869, opened 2015-04-10
Migrate embeds_many to has_many relation

ronin:
This PR changes deploys relation from embeds_many to has_many. This is proposed in this issue https://github.com/errbit/errbit/issues/868

### DIFF
--- a/app/models/app.rb
+++ b/app/models/app.rb
@@ -25,7 +25,7 @@ class App
 
 
   embeds_many :watchers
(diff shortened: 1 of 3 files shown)


stevecrozz:
The deployment tracking feature is on the chopping block with #913, so I'm going to close this feature and we can consider deployment tracking to be deprecated at this point. Please get in touch with me if you're really attached to this feature because there's still some time to discuss it.

Issue #782, opened 2015-01-05
document method for adding issue trackers


stevecrozz-:
Expand plugin documentation especially around issue trackers. This should include details with examples at errbit/errbit_plugin and a summary in the README file at errbit/errbit with a pointer to errbit/errbit_plugin

arthurnn:
How about adding all these to the WIKI? I think errbit core README is already out of hand... We should start moving things outside of it, WDYT?

stevecrozz:
Yeah, we could definitely use a cleanup in the errbit/errbit README. I think we just need a note about plugins that says something like: 'Errbit can be extended with new functionality including third-party integrations using plugins' with a link to 'errbit/plugin' where the details are explained.

stevecrozz:
From #780, include information on how to update the errbit Gemfile to include plugins.

nofxx:
@arthurnn great to know I'm not the only one with some evals on Gemfile's hehe.

errbit doesn't try to be user friendly (on install), as it put plainly on the Readme,
so the fork looks the simplest solution.

Just to discuss this bundle feature called 'lock' with you guys:

Say or point me to whatever I still think the (Gemfile) lock shouldn't even exist.
If you need something in version X, the Gemfile will handle that. 
If you don't lock it should work up to the latest version.
Locking what's not locked in my Gemfile just creates confusion and possibly lost of my time.
So a add plugins in config.yml, eval to gem and create the lock at deploy time... what can go wrong? 
Heh.

arthurnn:
what we could do, so people can add their own plugins, is kinda what readmine does https://github.com/edavis10/redmine/blob/master/Gemfile#L106-L109 .  with a plugins folder.

stevecrozz:
It's true. I've been mulling over all the ideas I can think of. I think the ideal and most safe solution would be one where the plugins live in an entirely separate runtime like a separate web service, an embedded and isolated runtime like therubyracer, or just a separate worker process. But all of those solutions suffer from increased complexity.

Embedding plugins within the main errbit process provides the ultimate flexibility and doesn't complicate deployment. So it seems like the right tradeoff. But it also kind of stinks that plugins can bring in their own possibly conflicting dependencies. It removes our ability to do consistent, repeatable deploys and we can kiss our Gemfile.lock goodbye along with `bundle install --deployment` when users can arbitrarily add gem dependencies at runtime.

Let me know if you have any other ideas, but so far something along the lines of redmine's approach seems best.

[...]

PR #871, opened 2015-04-17
#782 add docs for installing tracker plugins

stevecrozz:
Here's my attempt at documenting the method for creating and/or installing issue tracker plugins. I removed most of the docs for configuring the individual issue trackers because firstly, I know that they're out of date and don't actually work with current versions of errbit, but secondly because those docs belong in the individual plugin repos.

Keep in mind this pattern is the blueprint for what we want to happen next for notification integrations as plugins. If there's anything we don't like, we should try to spot it now.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -107,3 +107,6 @@ gem 'uglifier'
 gem 'jquery-rails', '~> 2.1.4'
 gem 'pjax_rails'
 gem 'underscore-rails'
+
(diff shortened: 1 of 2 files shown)


stevecrozz:
I think here's a fundamental flaw in my plan. I'm going to have to rethink this a bit, but I'll leave this open in case there's any comments.

Issue #479, opened 2013-05-10
Document API


domurtag:
According to this [Google Groups post](https://groups.google.com/forum/?fromgroups=#!topic/errbit/LpXab5JFPPY), Errbit has an [API](https://github.com/errbit/errbit/tree/master/app/controllers/api/v1) for reading data back out of Errbit.

There are no docs for this API, and apparently it's not the same as the Airbrake API, so the only way to figure out how to use it is to read the source code.

tjnet:
I want to contribute this issue.

PR #872, opened 2015-04-18
add docs for statistics API

tjnet:
#### What's this PR do?

Here's my attempt at documenting the method for representing error statistics.

#### Where should the reviewer start?

Add new api docs in docs/api/stats.md

#### Any background context you want to provide?

There are no docs for this API, and apparently it's not the same as the Airbrake API, so the only way to figure out how to use it is to read the source code.
Documentation will help us to understand how we can get error statistics on errbit. :smile_cat: 

#### What are the relevant tickets?

https://github.com/errbit/errbit/issues/479

### DIFF
--- a/docs/api/stats.md
+++ b/docs/api/stats.md
@@ -12,7 +12,7 @@ Example:
 
 ```sh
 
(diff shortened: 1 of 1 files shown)


stevecrozz:
Thanks!

tjnet:
Thanks for your review!! :smile:

PR #877, opened 2015-05-03
no need for wrapper gem, airbrake can easily be configured to send info ...

espen:
...about current user

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,6 +203,21 @@ This will ensure that your application stays up to date with any schema changes.
   another issue tracker integration, you may need to maintain it yourself. See
   (issue trackers)[#issue-trackers]
 
+## User information in error reports
+
+Errbit can now display information about the user who experienced an error.
+This gives you the ability to ask the user for more information,
+and let them know when you've fixed the bug.
+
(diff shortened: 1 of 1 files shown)


stevecrozz:
Interesting. It looks like you're right. I'll have to give it a shot and if it works the same as airbrake_user_attributes, then we can probably just remove this section since the airbrake gem docs cover this.

espen:
Can confirm this is working on my install (latest errbit master)

stevecrozz:
This makes sense. Thanks for checking that out. Since there's nothing our users need to do, I'm just going to remove this section from the docs.

espen:
I think it's good to mention it in the docs (like with my PR). Even though it is a part of the Airbrake gem it is also a feature in Errbit.

stevecrozz:
Ok. This sounds good, but I was too slow in merging your change. Can you rebase it onto master?

espen:
Ok, done.

stevecrozz:
did you push it?

stevecrozz:
ah there it is, thanks @espen

PR #889, opened 2015-05-19
Adds in formatting for Slack

veverkap:
This adds in attachment formatting for Slack

### DIFF
--- a/app/models/notification_services/slack_service.rb
+++ b/app/models/notification_services/slack_service.rb
@@ -18,7 +18,47 @@ class NotificationServices::SlackService < NotificationService
   end
 
   def post_payload(problem)
-    {:text => message_for_slack(problem) }.to_json
+    {
+      :attachments => [
+        {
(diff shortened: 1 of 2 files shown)


stevecrozz:
looks nice. but it looks like you may have removed the one and only call to 'message_for_slack' so can you delete the method definition too?

veverkap:
@stevecrozz It was left in as a convenience method.  Would you rather have this?

```ruby
  def post_payload(problem)
    {
      :attachments => [
        {
           :fallback => "[#{problem.app.name}][#{problem.environment}][#{problem.where}]: #{problem.error_class} #{problem_url(problem)}",
           :pretext => "<#{problem_url(problem)}|Errbit - #{problem.app.name}: #{problem.error_class}>",
           :color => "#D00000",
           :fields => [ SNIP
```

stevecrozz:
I see. I just misread it. This looks fine to me. Thanks!

PR #892, opened 2015-05-22
Set special password for using in Teatro

divineforest:
This config changes admin's password to `TeatroTeatro` so Teatro can be used again :)


stevecrozz:
Nice idea

PR #893, opened 2015-05-27
Support HipChat v2 API

bdewater:
HipChat has a [new API](https://www.hipchat.com/docs/apiv2) for a while now, and the new API allows users to use their own API token instead of requiring an administrator to generate one for them. V2 is also the only available API version for people running their own HipChat server.

This PR updates the HipChat gem, uses a migration to set already configured HipChat notifications to v1, and allows users to switch to v2 at their own pace.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -130,8 +130,9 @@ GEM
       tilt
     hashie (2.0.5)
     hike (1.2.3)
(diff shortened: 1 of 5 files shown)


stevecrozz:
Nice work. I appreciate the attention to our existing users who may not immediately want to change their config.

PR #896, opened 2015-06-04
Add deploy "message" to e-mail notifications

warrenguy:
It's currently in the web UI but not included in the e-mail notifications.


stevecrozz:
Thanks

PR #914, opened 2015-07-24
Truncate long problem message, fixed #848

### LINKED ISSUE #848: Some errors not registered ("key too large to index")
I was notified about an error occuring in our app, but not by Errbit. When investigating, Errbit had created a few errors in `Self.Errbit`, complaining that the key is too large to index.

```Moped::Errors::OperationFailure: The operation: #<Moped::Protocol::Command @length=70 @request_id=20589 @response_to=0 @op_code=2004 @flags=[] @full_collection_name="errbit.$cmd" @skip=0 @limit=-1 @selector={:getlasterror=>1, :w=>1} @fields=nil> failed with error 17280: "Btree::insert: key too large to index, failing errbit.problems.$message_1 4740 { : \"Sass::SyntaxError: File to import not found or unreadable: basic-meta/.\nLoad paths:\n /tmp/d20150310-23453-1wcb67\n /home/railsapp-staging/app/a...\" }" See https://github.com/mongodb/mongo/blob/master/docs/errors.md for details about this error.```

Using Errbit from commit 6cb19c235400ed6ed566078f7afdcaabac0e28dd

The message part of the XML sent is 4611 bytes long.

vlmonk:


stevecrozz:
I like the idea, but it seems the specs are broken

PR #888, opened 2015-05-19
Heroku button

3zcurdia:

PR #916, opened 2015-08-04
heroku button: enforce ssl by default

linki:
https://github.com/errbit/errbit/pull/888 works perfectly, thx @3zcurdia. though, i would recommend enforcing ssl by defaul.


stevecrozz:
makes sense to me

PR #917, opened 2015-08-11
Mongoid5 and performance improvements

stevecrozz:
Upgrade to mongoid5 beta and many performance improvements included. The specs pass, but I have made a lot of changes here, so this could be a little rocky. I think the performance gains will be worth the trouble and I'm documenting them now.


stevecrozz:
Looks like there are two issues:
1) travis somwhow doesn't support any mongo other than 2.4
2) logging isn't working quite right

So this is not ready to merge, but may still be useful for review while I figure this out

PR #924, opened 2015-08-21
sort apps by name

lucianosousa:


stevecrozz:
Is this intended as a performance improvement so that we can avoid loading all the apps just to sort them in ruby?

lucianosousa:
also, better to find projects in the list when we have a lot of them.

stevecrozz:
This makes sense. I just can't tell what 'all.sort' used to do. It looks like it would just use the default spaceship operator (<=>) to sort the mongoid documents which seems like a meaningless no-op at best.

lucianosousa:
I like this approach because it uses the mongoid api. cleaner and elegant result.
:)

if you want I can try this idea you have, but I really don't recommend.

stevecrozz:
No this is fine, I don't have an idea I was just trying to figure out the point of the existing code. But I believe it has no point.

lucianosousa:
:+1:

PR #929, opened 2015-08-28
  Wrap jabber notifications in a timeout block

funglaub:
We had a lot of trouble recently with https://github.com/xmpp4r/xmpp4r/issues/23, passenger reports lots of requests in the queue and all workers were dead (not responding) causing errors not being reported to errbit.

I isolated the issue and it looks like the xmpp4r deadlock is responsible for this. I wrapped all those calls into a simple `Timeout#timeout` block and now they are gone.


stevecrozz:
Timeout does worry me a bit, see http://www.mikeperham.com/2015/05/08/timeout-rubys-most-dangerous-api/ It sounds like some kind of underlying IO resource is should have a timeout set, but doesn't, and applying a timeout using Timeout could end up causing other problems when we start running multiple errbit threads and sharing connections.

I could still be persuaded to add this, though, because I'm planning to move all the notifier concerns into external Gems. We could fast-track that effort a bit if you're willing to help, especially if you want to also maintain the not-yet-existing errbit_gtalk_plugin.

funglaub:
I'm aware that Timeout::timeout might be dangerous, but xmpp4r/xmpp4r#23 does cause de facto problems in our setup. I agree with you that Timeout might not be the best solution in this matter, but I don't see any other options right now to deal with the issue.  

I'd happy to support your efforts in refactoring the whole notificatier concerns ;)

stevecrozz:
@funglaub Would you like to connect on https://gitter.im/errbit/errbit and see if we can come up with a plan?

stevecrozz:
@funglaub I would like to pull this in because I'm clearly not getting around to moving these issue trackers into external services. Can you rebase this PR into just one commit?

funglaub:
@stevecrozz I screwed it up, sorry. Please see #989.

PR #932, opened 2015-08-31
fix Bitbucket file URL

yhourdel:
Hi,

Bitbucket URLs have changed, here is a fix.
I've tried to upgrade `spec/helpers/backtrace_line_helper.rb` but:
* it seems to be ignored
* when I rename it to `spec/helpers/backtrace_line_helper_spec.rb` the current spec fails with
```
Failure/Error: Fabricate.build(:backtrace_line, :file => "[PROJECT_ROOT]/path/to/asset.js")
     Fabrication::UnknownFabricatorError:
       No Fabricator defined for 'backtrace_line'
```

If you want me to fill this spec, please fix current spec and I'll be happy to upgrade it.

### DIFF
--- a/app/decorators/backtrace_line_decorator.rb
+++ b/app/decorators/backtrace_line_decorator.rb
@@ -76,7 +76,7 @@ class BacktraceLineDecorator < Draper::Decorator
 
   def link_to_bitbucket(app, text = nil)
     return unless app.bitbucket_repo?
-    href = "%s#cl-%s" % [app.bitbucket_url_to_file(decorated_path + file_name), number]
+    href = "%s#%s-%s" % [app.bitbucket_url_to_file(decorated_path + file_name), file_name, number]
     h.link_to(text || file_name, href, :target => '_blank')
   end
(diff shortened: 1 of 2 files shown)


stevecrozz:
It looks like I failed to clean up this area during a refactor. We're using draper now for decorating backtrace lines and I believe the spot you want to fix is in app/decorators/backtrace_line_decorator.rb

I'll add a task for myself to clean up that helper.

yhourdel:
@stevecrozz do you need me to do something now ?

stevecrozz:
@yhourdel: It would be great if you could make that same change in app/decorators/backtrace_line_decorator.rb , verifity that it works, and potentially even add a spec. Are you up for it?

yhourdel:
@stevecrozz: I updated the decorator, but I'm having trouble with the spec. Can you help ?

stevecrozz:
Looks like @rposborne has a mergable PR (#944) for this same issue. I'm going to go with that one.

PR #937, opened 2015-09-04
fixes #936 Moved script/rails to bin/rails

### LINKED ISSUE #936: Heroku console bin/rails No such file or directory
For debug purposes, I want to access the console on heroku and I caught this behavior, I noticed that the errbit was upgraded to the newest rails versions and continues with 'script/rails' folder from older versions.

```
bruno_arueira  2.2.0p0 (expressio)  $ heroku run console --app expressio-errbit
Running `console` attached to terminal... up, run.9656
bash: bin/rails: No such file or directory
```

Thanks in advance,

brunoarueira:


stevecrozz:
Sounds good to me

Issue #884, opened 2015-05-11
configurable fingerprinting


stevecrozz:
Fingerprinting errors is currently pretty opaque. Errors are fingerprinted based on rules defined in fingerprinting classes. Changing the fingerprinting algorithm is a system-wide change and requires a new initializer in the best case and requires implementing a new class in the worst case. Further, the classes have strange names that have little to do with the factors that determine uniqueness.

I propose we make fingerprinting configurable. In my proposal, there would be a number of possible factors that could be chosen to be part of the fingerprint algorithm such as:
- last line of backtrace
- whole backtrace
- exception name
- exception message
- action name
- etc

~~The default would be made to match the current default algorithm. But in the edit screen for each app, admins could choose the factors that should determine uniqueness of an error for that app. We would store these rules on the app itself, so the performance penalty should be low since we already load the  app when inserting error notifications.~~

Just to descope this story a bit more into the realm of reality, let's start with system-wide configurable fingerprinting and add app-level configuration only if it becomes necessary. This should reduce the complexity considerably and still provide something with plenty of value.

bekicot:
+1

stevecrozz:
Fixed with #946

PR #946, opened 2015-09-13
configurable notice fingerprinting

stevecrozz:
This work-in-progress aims to address #884. Feedback is welcome. Still to do:
- [x] make sure task to regenerate fingerprints still works
- [x] restore the object id filter in error messages
- [x] write documentation
- [x] more specs


stevecrozz:
This is close. I need docs and an icon for the new tab. I think I'll divert my energy onto a separate effort to install font-awesome and replace most of our icons with fa.

stevecrozz:
This is looking good. Would anyone like to try this out or provide review/feedback?

xdmx:
I've upgraded to the current master and the last airbrake-js, but the custom fingerprints don't seem to work correctly. After the upgrade I ran `rake errbit:notice_refingerprint` and also after each time I enabled/disabled each options in the config, but the errors weren't splitted/merged as expected.

The only thing that changes it's the backtrace.. but it shouldn't use it since I've set to use 0 lines

errbit

errbit1

PR #951, opened 2015-09-24
fixing mongoid-rspec problem with bundle install and also update mongoid gem for compatibility

lucianosousa:
I just receive an error when I tried run bundle install.

```
$ bundle install
Fetching source index from https://rubygems.org/
Could not find mongoid-rspec-2.3.0.beta in any of the sources
```

I'm updating here the mongoid and mongoid-rspec to fix this error


stevecrozz:
Thanks. It looks like mongoid-rspec-2.3.0.beta was yanked and this is the correct fix.

PR #958, opened 2015-10-06
GitHub enterprise login

### LINKED ISSUE #957: Setting GITHUB_URL to a GitHub Enterprise URL doesn't work for logins
I tried setting `GITHUB_URL` to our internal GitHub Enterprise instance in my `.env` file and I was still redirected to `github.com` when I attempted to log in via GitHub.

Looking at `config/initializers/devise.rb`, it looks like there isn't any code to set `client_options` correctly as indicated here: https://github.com/intridea/omniauth-github#github-enterprise-usage

```ruby
provider :github, ENV['GITHUB_KEY'], ENV['GITHUB_SECRET'],
    {
      :client_options => {
        :site => 'https://github.YOURDOMAIN.com/api/v3',
        :authorize_url => 'https://github.YOURDOMAIN.com/login/oauth/authorize',
        :token_url => 'https://github.YOURDOMAIN.com/login/oauth/access_token',
      }
    }
```

Has anyone successfully used a GitHub Enterprise server for Errbit logins? I can provide a patch if this isn't just user error on my part.

md5:
This patch allows a GitHub Enterprise `GITHUB_URL` to be used for logins when `GITHUB_AUTHENTICATION=true`.

Previously, Errbit would attempt to log the user in against GitHub.com even if `GITHUB_URL` was set to a GitHub Enterprise URL.

Fixes #957


stevecrozz:
I put a PR into your PR to address a few concerns I had:
https://github.com/appropriate/errbit/pull/1

Can you test those out and respond if anything looks wrong? I don't have a github enterprise account, so the best I can do on my own is to make sure the specs pass.

After we get that straightened out, the last thing would be to squash all these commits into logical units (probably just one commit), to keep the git log navigable.

md5:
I made a comment about one minor issue. I'm fine with merging it without addressing that particular problem.

Happy to merge and squash it if you'd like to proceed with it as is.

stevecrozz:
Yeah, if that's the only issue, let's get it squashed and then I say its ready. Here's a placeholder for the issue you brought up: https://github.com/errbit/errbit/issues/959

md5:
@stevecrozz Merged your PR and squashed.

stevecrozz:
Excellent. I did see your comments about the two URLs and the possibility for getting those mixed up. I'm still thinking about that, but I'm inclined to remedy that with a nice documentation section on setting up errbit with github enterprise.

Thanks for your effort here. It's just in time for v0.5.0

md5:
> Excellent. I did see your comments about the two URLs and the possibility for getting those mixed up. I'm still thinking about that, but I'm inclined to remedy that with a nice documentation section on setting up errbit with github enterprise.

Since there couldn't have been anybody using an enterprise `GITHUB_URL` with `GITHUB_ORG_ID` and `GITHUB_AUTHENTICATION=true`, since it didn't work previously, it's probably a low risk to existing users. Documenting the proper configuration sounds fine.

PR #963, opened 2015-10-11
Install rubocop - auto-validating syntax checking

rud:
Hi,

I've added the "rubocop" gem to the test and development environments. I've then proceeded to fix all the "lint" style warnings I've seen, in a series of focused commits. These are all be extremely safe changes to make, and provides the foundation for further source-code style unifications.

What do you all think, is this valuable work?


stevecrozz:
Very good. I just have the same feedback here as https://github.com/errbit/errbit/pull/964

stevecrozz:
Looks good. Thanks for starting this effort!

rud:
Awesome, thank you for merging! :)

PR #965, opened 2015-10-11
RSpec: run in random order

rud:
This is usually a good way to ferret out state leaking from one test to another, making the tests frustrating to work with. Fortunately it seems like everything is in pretty good order. 

I've been running this in a console locally for a while:
```
 while (true) { bundle exec rspec --order random || say "failed once" }
```
Looks good so far. 

If this passes on Travis I'll push an update to use a random test ordering each time, and not a fixed seed like now.

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,2 +1,3 @@
 --color
 --require spec_helper
+--order random
--- a/spec/models/notification_service/notification_service_spec.rb
+++ b/spec/models/notification_service/notification_service_spec.rb
@@ -9,7 +9,7 @@ describe NotificationServices, type: 'model' do
   end
 
   it "it should use the protocol value specified in the config in #problem_url" do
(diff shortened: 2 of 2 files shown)


stevecrozz:
It's a good idea. Thanks! I see why you made the spec change, but actually this change negates the purpose of the test which is here to verify the default. I think instead of modifying this test, we should modify the other test with an allow(Errbit::Config).to_receive(:protocol).and_return('https') which I believe will leave the global variable unmodified and allow a real test of the default on line 8.

Also, if you could squash these commits into a single commit, that helps keep the git log easy to navigate.

rud:
Squashed and now using a mock instead of class variable modification.

As a side note, maybe it would be handy to have a "reset to default global config" method somewhere, especially useful in specs. Just an idea.

PR #972, opened 2015-10-15
Rubocop whitespace fixes

rud:
I'm sorry, seemed like a good idea when I started, but this may be an attempt at fixing entirely too much whitespace in the name of code style consistency. 

The Rubocop style TODO is still not empty, but it it steadily shrinking.


rud:
By adding `?w=0` to the diff view most of the whitespace changes can be ignored (making for a shorter diff). Heres a link with that: https://github.com/errbit/errbit/pull/972/files?w0

I expect the most controversial commit by far will be 832715efa0e61b77fbf82c4fbb72c1f4dd798808 - fortunately there are other styles to choose from as well, we can easily tweak it.

app/controllers/apps_controller.rb
```diff
@@ -90,60 +90,60 @@ def regenerate_api_key
...
+        params[:app][:notification_service_attributes][:notify_at_notices] = notify_at_notices
+      else
+        default_array = params[:app][:notification_service_attributes][:notify_at_notices] = Errbit::Config.notify_at_notices
+        flash[:error] = "Couldn't parse your notification frequency. Value was reset to default (#{default_array.join(', ')})."
       end
     end
+  end
 
-  private
+private
```
stevecrozz: My preference is to use the format:
```ruby
private def app_params
  params.require(:app).permit!
end
```

It supports only ruby 2.1+, but I do like it and I think rubocop will allow it by default. Any thoughts?

app/models/app.rb
```diff
@@ -19,9 +19,9 @@ class App
   # Some legacy apps may have string as key instead of BSON::ObjectID
   # identity :type => String
   field :_id,
-    type: String,
-    pre_processed: true,
-    default: ->{ BSON::ObjectId.new.to_s }
+        type: String,
+        pre_processed: true,
+        default: ->{ BSON::ObjectId.new.to_s }
```
stevecrozz: This is one cop where I've run into trouble before. Maybe I have to find a way to get used to this, but the previous incantation looks more natural to me.

app/models/issue_tracker.rb
```diff
@@ -15,19 +15,19 @@ def tracker
         klass = ErrbitPlugin::Registry.issue_trackers[type_tracker] || ErrbitPlugin::NoneIssueTracker
         # TODO: we need to find out a better way to pass those config to the issue tracker
         klass.new(options.merge(
-          github_repo: app.try(:github_repo),
-          bitbucket_repo: app.try(:bitbucket_repo)
+                    github_repo: app.try(:github_repo),
+                    bitbucket_repo: app.try(:bitbucket_repo)
```
stevecrozz: This cop doesn't make sense to me. The previous format is more legible to my eyes. Is there a way to enforce that format instead of this new one?

app/models/notification_services/gtalk_service.rb
```diff
@@ -1,37 +1,37 @@
...
 
   def check_params
-    if Fields.detect { |f| self[f[0]].blank? && self[f[2]].blank? } 
-      errors.add :base, 
-      """You must specify your Username, Password, service, service_url 
-         and either rooms or users to send to or both"""
+    if Fields.detect { |f| self[f[0]].blank? && self[f[2]].blank? }
+      errors.add :base,
+                 """You must specify your Username, Password, service, service_url
+                    and either rooms or users to send to or both"""
```
stevecrozz: This kind of thing looks so ugly to me, that I have tended to avoid it entirely by just putting the string into a class constant short enough to fit on one line.

app/models/problem.rb
```diff
@@ -85,25 +84,25 @@ def self.cache_notice(id, notice)
...
+                                                       "hosts.#{host_digest}.value" => notice.host,
+                                                       "user_agents.#{user_agent_digest}.value" => notice.user_agent_string,
+                                                     },
+                                                     '$inc' => {
+                                                       'notices_count' => 1,
+                                                       "messages.#{message_digest}.count" => 1,
+                                                       "hosts.#{host_digest}.count" => 1,
+                                                       "user_agents.#{user_agent_digest}.count" => 1,
+                                                     }
+                                                   }, return_document: :after)
```
stevecrozz: Are there any other options? This really ruins my 3-way diffs especially on small monitors and causes me to have to horizontally scroll to see the code.

config/initializers/devise.rb
```diff
@@ -247,9 +247,9 @@
     }
 
     config.omniauth :github,
-      Errbit::Config.github_client_id,
-      Errbit::Config.github_secret,
-      github_options
+                    Errbit::Config.github_client_id,
+                    Errbit::Config.github_secret,
+                    github_options
```
stevecrozz: This irks me, but I could probably learn to live with it. In general, I really prefer multi-line indentation to be the normal, 2-space style. I tend to make use of namespaces as well as descriptive variable names and its really easy to end up with super long lines. It hasn't happened in this case thankfully, but its very easy to end up there.

stevecrozz:
Most of this looks great. I got about halfway down and found myself just repeating the same one or two comments so I just stopped there because I don't think I have any other feedback.

app/models/app.rb
rud: You're right, this doesn't look too hot just yet.

There's a tweakable setting called `Style/AlignParameters`: https://github.com/bbatsov/rubocop/blob/2c9d207af5ffb639d32f17a97251956f16c2356b/config/default.yml#L135-L152 Your description matches the mode `with_fixed_indentation`, giving that a try.

rud: Fixed in 52d8b62cbc0

app/controllers/apps_controller.rb
rud: I can certainly try and get rubocop to play along. That said, the keyword does become kind of hidden in this compact style, where access modifiers are otherwise used to demarcate sections of code with different audience scope: public for the world, private/protected for internal details. By putting it on a line with a method definition, to me it communicates that it affects only this one method, but I'm sure if you add another method definition below this, it would also be `private`. That probably means the line `private def app_params` should be broken up when adding to the code, creating an extra burden of consistency on contributors.

What do you think, do we look for alternatives to what is currently here?

app/models/issue_tracker.rb
rud: Yes, fixed with 52d8b62cbc0b6733bf5c00bad939d3e2be9c352d

app/models/notification_services/gtalk_service.rb
rud: Have no fear, 52d8b62cbc0b6733bf5c00bad939d3e2be9c352d brings it back to order

app/models/problem.rb
rud: Good catch - fixed in a216c43 by putting the first argument on the next line.

config/initializers/devise.rb
rud: The style-switch in 52d8b62cbc0b6733bf5c00bad939d3e2be9c352d also fixes this case, thankfully.

rud:
@stevecrozz excellent - and I think I've been able to address most of your comments right away, primarily with 52d8b62cbc0b6733bf5c00bad939d3e2be9c352d. We still need to talk keywords, though. Let me hear what you think of https://github.com/errbit/errbit/pull/972#discussion_r42209340

app/controllers/apps_controller.rb
stevecrozz: Interestingly when you put it on one line the way I have, it does actually affect only that method. It works because in Ruby 2.1+, method definitions return their own name as a symbol and at that point private is just a method taking an argument. Since 'def app_params...' returns :app_params, the expression becomes `private :app_params` after the method definition is evaluated.

app/models/problem.rb
```diff
@@ -43,12 +42,12 @@ class Problem
...
-    message: "text",
-    app_name: "text",
-    environment: "text"
-  }, default_language: "english")
+          error_class: "text",
+          where: "text",
+          message: "text",
+          app_name: "text",
+          environment: "text"
+        }, default_language: "english")
```
stevecrozz: Can this remain the way it was?

config/mongo.rb
```diff
@@ -11,13 +11,13 @@
...
-  })
+                              clients: {
+                                default: {
+                                  uri: uri
+                                }
+                              },
+                              options: {
+                                use_activesupport_time_zone: true
+                              }
+                            })
```
stevecrozz: And this?

lib/tasks/errbit/demo.rake
```diff
@@ -43,28 +43,27 @@ namespace :errbit do
...
+                                                       },
+                                                       :server_environment => {'environment-name' => Rails.env.to_s},
+                                                       :notifier => {:name => "seeds.rb"},
+                                                       :app_user => {
+                                                         :id => "1234",
+                                                         :username => "jsmith",
+                                                         :name => "John Smith",
+                                                         :url => "http://www.example.com/users/jsmith"
+                                                       }
+                                                     })).generate_notice!
```
stevecrozz: And this

spec/controllers/api/v3/notices_controller_spec.rb
```diff
@@ -16,9 +16,9 @@
     post :create, legit_body, legit_params
     notice = Notice.last
     expect(JSON.parse(response.body)).to eq({
-      'id' => notice.id.to_s,
-      'url' => app_problem_url(app, notice.problem)
-    })
+                                              'id' => notice.id.to_s,
+                                              'url' => app_problem_url(app, notice.problem)
+                                            })
```
stevecrozz: There's a bunch more instances of this specific thing in the ./spec folder. If these multi-line hash literals can be just 2-space indented, then I think this PR would be ready to go.

app/controllers/apps_controller.rb
rud: Oh that is exceedingly clever, I did not know that. I'll look into what options are available, might take a bit, though.

rud: Fixed this manually in 3c4fad07af186fefffe7ee2b463f46eae7484088, rubocop doesn't care. Yay, think I'm through your comments. What do you think?

stevecrozz:
Looks great so far. There's a number of hyper-indented hash literals in the spec folder. I don't mind pointing them all out or fixing them myself. You've been a great help already, but I'm assuming you have the stamina to follow this through since you've shown that already.

There is also one failing test which I haven't bothered to figure out. Let me know if you need any help on that.

rud:
I think the winning strategy here will be to break this down into smaller bits that are trivially reviewable and mergeable. Going slow and steady is probably going to win the day here.

rud:
See also #980

rud:
Expecting to circle back to this at some point, but if anyone else feels like resurrecting the still unmerged commits, do feel free. Alternatively, simply whittling down the `.rubocop_todo.yml` file would also be a neat plan.

PR #977, opened 2015-10-19
Adds "Initial setup" to heroku deploy doc

### LINKED ISSUE #971: Default login for errbit running on Heroku?
What is the default admin login for deployments made using the heroku button located at https://github.com/errbit/errbit/blob/master/docs/deployment/heroku.md ?

sbruhns:
Adds commands needed to setup a new errbit app to the "The Easy Way" section.

This might be helpful for new users and resolve #971.

### DIFF
--- a/docs/deployment/heroku.md
+++ b/docs/deployment/heroku.md
@@ -7,6 +7,28 @@ you can use this deploy button to get a basic deployment running on Heroku.
 
 [![Deploy](https://www.herokucdn.com/deploy/button.svg)](https://heroku.com/deploy?template=https://github.com/errbit/errbit/tree/master)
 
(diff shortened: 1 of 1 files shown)


stevecrozz:
Cloning the repo shouldn't be necessary because you can provide the app name to the heroku binary with --app. And I'm thinking we can add 'heroku run rake errbit:bootstrap' to app.json under scripts.postdeploy so it runs automatically. The heroku config is optional, but probably valuable since emails and links within emails won't point to the correct URLs without it, but maybe that can also be inferred from the heroku environment.

I would really like it if the easy way was a single step, click the button. Let's try to push the app in that direction.

PR #979, opened 2015-10-19
Chore/rubocop newline fixes

rud:
Focusing on places with too many or too few newlines. Does this all look good to you?


rud:
Splitting #972 into smaller chunks so we can get them right.

stevecrozz:
lgtm. thanks.

PR #980, opened 2015-10-19
Rubocop: consistent access modifier indentation

rud:


rud:
Builds on #979, extracted from #972

PR #983, opened 2015-10-26
fixes problem in Issue#982 

meh-:
fixes problem in Issue#982 "This error has occured 0 times" instead of 1 time.

### DIFF
--- a/app/mailers/mailer.rb
+++ b/app/mailers/mailer.rb
@@ -15,8 +15,9 @@ class Mailer < ActionMailer::Base
   def err_notification(error_report)
     @notice   = NoticeDecorator.new error_report.notice
     @app      = AppDecorator.new error_report.app
+    @problem  = ProblemDecorator.new error_report.problem
 
-    count = error_report.problem.notices_count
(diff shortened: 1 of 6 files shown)


stevecrozz:
Thanks @meh- 

This is probably the right place to make a change, but theoretically it should not be necessary to reload the document because we already return it in Problem.cache_notice (https://github.com/errbit/errbit/blob/1ad440bd3bedc406029478c2201d61316172925c/app/models/problem.rb#L105)

It would be good to avoid this extra database query for performance reasons.

meh-:
ok, may be something like 
@problem = Problem.cache_notice(@error.problem_id, @notice)
@notice.problem = @problem

stevecrozz:
Yeah that could work, but if possible I think it would be good to avoid that second line and stop accessing @notice.problem wherever we're doing that. Instead access @problem directly.

stevecrozz:
Thanks @meh- I was coming back to this just now and noticed the issue is no longer present. I think some other fix must have taken care of this problem.

PR #985, opened 2015-10-28
fixed button 'show/hide resolved' on errors#index page

meh-:
Fixed bug:
1. Go to errors page /problems
2. Title = Unresolved Errors, button = show resolved
3. Click show resolved button (link actually)
4. Title has the same name instead of 'All Errors' and button is the same as before. So now we can't hide resolved errors

Problem was in using instance variable inside view.


stevecrozz:
Thanks @meh-

Issue #926, opened 2015-08-26
NameError: uninitialized constant IssueTrackers


kubaaa:
I update Errbit to last versions. If i click link http://errbit.net/apps, i get error  NameError: uninitialized constant IssueTrackers

activesupport-4.1.12/lib/active_support/inflector/methods.rbmethods.rb:238→ const_get
activesupport-4.1.12/lib/active_support/inflector/methods.rbmethods.rb:238→ block in constantize
activesupport-4.1.12/lib/active_support/inflector/methods.rbmethods.rb:236→ each
activesupport-4.1.12/lib/active_support/inflector/methods.rbmethods.rb:236→ inject
...
mongoid-5.0.0.beta/lib/mongoid/relations/accessors.rbaccessors.rb:100→ block in get_relation
mongoid-5.0.0.beta/lib/mongoid/threaded/lifecycle.rblifecycle.rb:89→ _building
mongoid-5.0.0.beta/lib/mongoid/relations/accessors.rbaccessors.rb:99→ get_relation
mongoid-5.0.0.beta/lib/mongoid/relations/accessors.rbaccessors.rb:187→ block in getter
app/models/app.rbapp.rb:129→ issue_tracker_configured?
draper-2.1.0/lib/draper/delegation.rbdelegation.rb:10→ issue_tracker_configured?
app/helpers/apps_helper.rbapps_helper.rb:47→ block in detect_any_apps_with_attributes
app/helpers/apps_helper.rbapps_helper.rb:44→ each
app/helpers/apps_helper.rbapps_helper.rb:44→ detect_any_apps_with_attributes
app/helpers/apps_helper.rbapps_helper.rb:15→ any_github_repos?
app/views/apps/index.html.hamlindex.html.haml:9→ _app_views_apps_index_html_haml__2492920191333893560_70247947053460

haobole:
:+1:

stevecrozz:
Thanks for the report @kubaaa . This could be due to a recent change I made in cleaning out a bunch of old migrations. Can you verify that you have followed these instructions?
https://github.com/errbit/errbit#upgrading-errbit-beyond-v040

kubaaa:
Yes. I run rake db:migrate on server. I use Redmine Tracker in my server.  The server has created an early app. If I try to view the app or to edit, I get the above error. To deploy, use this manual https://github.com/errbit/errbit/blob/master/docs/deployment/capistrano.md

stevecrozz:
[...]

PR #986, opened 2015-10-29
Restore migration to extract issue trackers

jeremywadsack:
Addresses #926.

Between 0.4.0 and 0.5.0 all old migrations were removed because they were assumed to have been run in any production environment. However, it appears that the data migration for the IssueTrackers collection was [changed](https://github.com/errbit/errbit/commit/8dbfda4566fbf24b73f6173967fb0d0de032d296) during the process of building 0.4.0. Which meant anyone before then would not have the new structure in place. The proposed solution of migrating to 0.4.0 and then migrating forwards does not work if the MongoDB database underlying errbit was already moved to MongoDB 3.0 (which MongoLabs did for all their databases), because 0.4.0 doesn't have the updated Mongoid 5 drivers needed to connect.

This restores the commit so that it errbit installations before 0.4.0 can be upgraded directly to 0.5.0 using the new Mongoid drivers. Following @jerefrer's suggestion from https://gitter.im/errbit/errbit I patched the migration to work with the new Mongoid drivers and tested in our production setting.

### DIFF
--- a/db/migrate/201510290041_extract_issue_tracker.rb
+++ b/db/migrate/201510290041_extract_issue_tracker.rb
@@ -1,0 +1,46 @@
+class ExtractIssueTracker < Mongoid::Migration
+
+  TRACKER_MAPPING = {
+    'ErrbitTracPlugin::IssueTracker' => 'trac',
+    'IssueTrackers::BitbucketIssuesTracker' => 'bitbucket',
+    'IssueTrackers::FogbugzTracker' => 'fogbugz',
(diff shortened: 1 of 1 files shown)


stevecrozz:
This looks great @jeremywadsack! There is a very trivial build failure because we're using rubocop now. If you fix it, this should be ready to go.

stevecrozz:
Excellent. Hopefully this will keep folks from running into this problem. Thanks for the contribution.

PR #989, opened 2015-11-03
  Wrap jabber notifications in a timeout block

funglaub:
I screwed it up in #929


stevecrozz:
Looks like there's a conflict. I might have just caused that by merging another PR. Sorry if that's the case. Can you update it once more?

funglaub:
@stevecrozz #997

PR #990, opened 2015-11-03
add unicorn worker killer

8398a7:
I want to make savings of memory.
How about that?


stevecrozz:
I think its a good idea, but I would like to switch from unicorn to puma. So this would be nice for now, but ultimately could be a wasted effort once we switch. If you're ok with that, then we can go ahead and merge this.

But before that, I think it would be good to keep the existing behavior. In other words, let's not enable the killers unless the user has set request / memory thresholds. Then we should change the names of these methods to be more generic like KILL_ON_REQUEST_COUNT_MIN and KILL_ON_RSS_MIN.

And once that's ready, we need to have the new parameters documented in https://github.com/errbit/errbit/blob/master/docs/configuration.md.

8398a7:
Thank you for feedback.
I will try implementation.

8398a7:
@stevecrozz I fixed.

stevecrozz:
thanks. merged

PR #991, opened 2015-11-04
CI: enable rubocop checks

rud:


rud:
This seems like a random failing - is this a known issue?

rud:
Will add this back at a later date

PR #995, opened 2015-11-10
Errbit capistrano/unicorn

7Dump:


stevecrozz:
I'm not sure what I'm looking at. Can you explain this?

7Dump:
@stevecrozz I'm sorry it's my fault, we use errbit and I mistakenly added changes to your branch.

PR #1000, opened 2015-11-16
Fix syanx error

nacengineer:


stevecrozz:
Thanks. But this is not a syntax error if you use ruby 2.1 or greater and errbit does require ruby 2.1+

PR #1002, opened 2015-12-02
Update gem uglifier in order to avoid security vuln as reported by gemnasium

compwron:
https://gemnasium.com/errbit/errbit 
screen shot 2015-12-01 at 5 12 07 pm

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -124,7 +124,7 @@ GEM
       octokit
     errbit_plugin (0.5.0)
     erubis (2.7.0)
-    execjs (2.5.2)
+    execjs (2.6.0)
     fabrication (2.13.2)
     faraday (0.9.1)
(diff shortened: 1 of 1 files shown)


stevecrozz:
Why did you include a change to .gitignore?

compwron:
In order to ignore the bin directory, where binstubs are stored, which do not appear to have been  previously tracked in version control. https://robots.thoughtbot.com/use-bundlers-binstubs

stevecrozz:
Can you exclude that change from the PR? It seems like the Rails recommendation is to add any binstubs you find useful to the ./bin folder and check them into source control:
http://api.rubyonrails.org/v4.2.0/classes/Rails/AppRailsLoader.html

compwron:
Ok I took it out :)

stevecrozz:
Thanks for keeping an eye on the Gemfile!

Issue #981, opened 2015-10-20
Password reset broken


robinboening:
Hey there,

Situation: (Errbit v0.5.0) I request a new password and click on the link from the mail to set a new password. Then I type a new password and submit the form. 

Issue: I receive an error telling me I need to provide a username.

Expected behavior: After submitting the form with the new password, it should be set (without the need of a username).

<img width="679" alt="errbit_password-reset" src="https://cloud.githubusercontent.com/assets/342977/10607548/858efd32-773a-11e5-9316-db6a3a72338d.png">

stevecrozz:
It seems that you are running errbit with ERRBIT_USER_HAS_USERNAME=true, but your user does not have a username. When this flag is enabled, it means users without usernames are invalid. What's interesting about this is you shouldn't be able to have a user in this state unless you were formerly running with ERRBIT_USER_HAS_USERNAME=false and enabled it later. Is that what's going on?

stevecrozz:
My plan is to disable all validations other than password validations when doing a password reset. This should be safe enough since only the password can be updated during a password reset.

yannvery:
Do you need a PR for this one ?

stevecrozz:
@yannvery that would be great

PR #1003, opened 2015-12-04
Disable user validation when reset password

yannvery:
Reference #981 
I don't know if it's the best approach. What do you think?

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -81,6 +81,15 @@ class User
     :auth_token
   end
 
+  def reset_password(new_password, new_password_confirmation)
+    self.password = new_password
+    self.password_confirmation = new_password_confirmation
+
+    self.class.validators_on(:password).map { |v| v.validate_each(self, :password, password) }
+    return false if errors.any?
+    save(validate: false)
+  end
+
   private def generate_authentication_token
     loop do
       token = Devise.friendly_token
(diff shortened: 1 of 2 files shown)


app/models/user.rb
```diff
@@ -81,6 +81,19 @@ def self.token_authentication_key
     :auth_token
   end
 
+  def reset_password(new_password, new_password_confirmation)
```
stevecrozz: I haven't tested this much, but how about something like:
```ruby
self.class.validators_on(:password).map { |v| v.validate_each(self, :password, self.password) }
return false if user.errors.any?
save(validate: false)
```

It doesn't seem like there are any validators installed on password_confirmation, but maybe there's some magic that I'm not aware of.

spec/models/user_spec.rb
```diff
@@ -35,6 +35,12 @@
       user2.save
       expect(user2).to be_valid
     end
+
+    it "disables validations when reset password" do
+      user = Fabricate.build(:user, email: '')
+      user.save(validate: false)
+      expect(user.reset_password('Password123', 'Password123')).to be_truthy
+    end
   end
```
stevecrozz: Would be nice to have a negative test case, to make sure password validations still prevent you from saving a password that is too short for example when using User#reset_password.

also reset_password is fine, but I might go with set_password because reset brings to my mind a slightly different idea of what this is doing.

yannvery:
@stevecrozz I made changes after your notes.
I used 'User#reset_password' to override Devise method, do you prefer an alias method named 'set_password' on this one?

stevecrozz:
It looks great. Thanks for this @yannvery !

Issue #999, opened 2015-11-15
change server support from unicorn to puma


stevecrozz:
Unicorn has been great, but puma has a cleaner API and easier threading support. According to anecdotal evidence from various blog entries I've found, puma has significant performance advantages and is also now [officially recommended over unicorn by heroku](https://devcenter.heroku.com/changelog-items/594).

otzy007:
:+1:

stevecrozz:
Fixed in master

Issue #978, opened 2015-10-19
How to configure Errbit to fix "Errbit Timeout::Error: Waited 0.5 sec"


taelor:
Does anyone have any suggestions on how to configure errbit to avoid this error? I'm not sure what exactly the problem is, but I have a feeling it might be our Mongo configuration. I'm not too familiar with Mongo, so I'm not sure what I would need to change to help avoid this issue.

Is there some sort of connection pool that I might need to turn up on Mongo? 

```
connection_pool-2.1.1/lib/connection_pool/timed_stack.rb:85→ block (2 levels) in pop
connection_pool-2.1.1/lib/connection_pool/timed_stack.rb:77→ loop
connection_pool-2.1.1/lib/connection_pool/timed_stack.rb:77→ block in pop
connection_pool-2.1.1/lib/connection_pool/timed_stack.rb:76→ synchronize
...
mongoid-4.0.2/lib/mongoid/contextual/mongo.rb:198→ block in first
mongoid-4.0.2/lib/mongoid/contextual/mongo.rb:457→ try_cache
mongoid-4.0.2/lib/mongoid/contextual/mongo.rb:197→ first
mongoid-4.0.2/lib/mongoid/contextual.rb:20→ first
app/models/error_report.rb:39→ app
app/models/error_report.rb:78→ valid?
app/controllers/notices_controller.rb:13→ create
actionpack-4.1.8/lib/action_controller/metal/implicit_render.rb:4→ send_action
actionpack-4.1.8/lib/abstract_controller/base.rb:189→ process_action
actionpack-4.1.8/lib/action_controller/metal/rendering.rb:10→ process_action
actionpack-4.1.8/lib/abstract_controller/callbacks.rb:20→ block in process_action
...
puma-2.11.2/lib/puma/server.rb:375→ process_client
puma-2.11.2/lib/puma/server.rb:262→ block in run
puma-2.11.2/lib/puma/thread_pool.rb:104→ call
puma-2.11.2/lib/puma/thread_pool.rb:104→ block in spawn_thread
```

stevecrozz:
@taelor are you using the google talk / jabber / xmpp notifier?

taelor:
No, not using either of those. It also seems like its happening before the error is getting created/logged.

stevecrozz:
@taelor You may be running out of connections in your connection pool. Most errbit users likely use a single-threaded, multiple-process unicorn server and that's the way we have documented the capistrano and heroku deployment scenarios.

[...]

PR #1005, opened 2015-12-14
Use puma instead of unicorn

otzy007:
unicorn gem and configs have been replaced with puma configs found on https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server

I've removed unicorn from documentation too and the capistrano deploy script has been updated to deploy with puma using the capistrano3-puma gem.

I've tested the deployment on heroku and with capistrano on a ruby:2.2 docker container.

This is a PR for #999


stevecrozz:
Looks great! it's worth noting that this change not only replaces unicorn with puma, but also changes the application container model from pure workers to workers + threads. Having the ability to do this easily is one of the reasons I like puma, but it also has the potential to expose issues we haven't yet had to deal with such as https://github.com/errbit/errbit/issues/978

We can deal with those if they appear. I think its time.

stevecrozz:
@otzy007 Would you mind updating the docs in ./docs/deployment/capistrano.md to reflect your changes?

stevecrozz:
One other thing is the unicorn_worker_killer. We still have configuration in .env.default such as KILL_ON_RSS_MIN=250 which is useful for killing unicorn processes that exceed certain resource limitations. If we can maintain this functionality, that would be awesome. Otherwise we should dump these configuration parameters.

otzy007:
@stevecrozz ok. I'll take care of the capistrano documentation and I'll check if puma needs or has some kill options.

otzy007:
I've updated the docs and the capistrano template.
About the KILL_ON_RSS_MIN, puma doesn't have such a feature. Correct me if I'm wrong please.
It has just this timeout config https://github.com/puma/puma/blob/master/examples/config.rb#L155

From my past experience with puma I've been monitoring the process with monit to check if it has crashed or it's exceeding an imposed limit of RAM.

So. Should I dump those KILL_ON_RSS ?

stevecrozz:
I think you're right @otzy007. And I think these KILL_ON params all need to go. If you take a look through https://github.com/errbit/errbit/pull/990 you should be able to spot all the changes. It'll be a minor bummer for @8398a7, but I think the benefits of puma outweigh the lack of worker killer.

Maybe we can consider https://github.com/xeger/stalin or link to some docs on setting up monit in the future.

otzy007:
Monit is pretty straight forward. There's a capistrano task to upload the config. You just have to edit the config if you want to limit the memory usage.

Stalin looks good too, but what if puma crashes? Monit will restart it, but stalin is unable. In the past for me puma crashed a few times on rubinius or jruby

otzy007:
Here are the docs for monit 😎 d902148e1ba87e2c6f2f908993348965c9b88a42

stevecrozz:
Thanks @otzy007. This is looking really good!

I have a couple last questions/notes:
- Why port 3000? It looks like in the past we used 8080 and I don't see the reason for the change
- puma can live in the regular non-production section of the Gemfile unless you see a reason it shouldn't. I plan to run puma for both development and production
- should we consider using the default capistrano-puma puma.rb template for capistrano deployments? https://github.com/seuros/capistrano-puma/blob/master/lib/capistrano/templates/puma.rb.erb (can be generated as part of the `rails generate capistrano:nginx_puma:config` generator)

otzy007:
1. Ok. I'm changing to 8080. 3000 was into the example config from heroku that I've copy-pasted mostly
2. OK :) Actually even if you set  puma under the production group `rails s` will start with puma
3. Yes.  puma.rb config is generated by capistrano based on the settings on deploy.rb for capistrano deployments. I've just forgot to remove the symlink.

One more thing. With the capistrano deployment puma is started without workers using the default setting and with the default number of threads (0, 16). On JRuby workers are not available.

stevecrozz:
It makes sense that workers are not available in JRuby. Since JRuby can make much better use of multiple cores, there's a lot less reason to run multiple workers. I'm good with the capistrano-puma defaults because I'm sure users will end up customizing config/deploy.rb to suit their own needs.

This seems to be ready to go. I'm thinking of releasing one final version of Errbit before merging this PR because this is a substantial change for those of us who are used to deploying unicorn.

otzy007:
Great. Thank you 😀

stevecrozz:
@otzy007 I'm prepping the v0.6.0 release and this is the first thing I plan to merge after that. But the PR has acquired conflicts. Can you resolve them over the next day or two? If not, I can probably tackle it too.

otzy007:
Sure. I'll rebase it tomorrow

otzy007:
it's ready to be merged now 😃

stevecrozz:
done. thanks once again @otzy007 !

otzy007:
I'm glad I can help 😃

PR #1014, opened 2015-12-22
Add syslog functionality

crunis:
@stevecrozz this adds the syslog functionality but something seems wrong since some output is still replicated in STDOUT. You may want to take a look.

### DIFF
--- a/config/environment.rb
+++ b/config/environment.rb
@@ -1,5 +1,3 @@
-require 'syslog/logger'
-
 # Load the Rails application.
 require File.expand_path('../application', __FILE__)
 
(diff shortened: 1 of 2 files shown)


stevecrozz:
This looks good. Can you move the require statement inside the elsif block so that file is only required when needed?

Then it's probably time to document this parameter. Would you mind adding docs on ERRBIT_LOG_LOCATION to ./docs/configuration.md?

crunis:
@stevecrozz : Done !

stevecrozz:
Almost ready. The default is STDOUT, not no logging. The defaults are all here btw: https://github.com/errbit/errbit/blob/master/.env.default

crunis:
fixed

stevecrozz:
looks good. thanks for the contribution!

Issue #973, opened 2015-10-16
App sorting seems incorrect


funglaub:
I'm using errbit from the current master and the sorting of all apps by the unresolved count seems to be wrong: apps with the highest unresolved count appear somewhere in the middle or at the end of the list. The first item has an unresolved error count of 0.

Is this a bug or did I miss something when upgrading?

stevecrozz:
Yeah this was my mistake and I just realized it. We had another user @lucianosousa who wanted apps to be sorted differently: https://github.com/errbit/errbit/pull/924

And I didn't understand the intent at the time, but I ran across the answer it a few days ago in how it used to work: https://github.com/errbit/errbit/blob/v0.4.0/app/models/app.rb#L179-L184

We apparently have conflicting desires about how apps should be sorted and I don't really want to introduce fancy per-column sorting at this time. I'm inclined to revert #924 now that I understand what happened.

On the other hand, this is the first complaint I've heard about the new sorting method. Is this new method a problem for you @funglaub ?

lucianosousa:
what I did was just sort by name.
my problem before it was:
I've a lot of apps with different envs, eg: [ app acceptance, app staging, app production]

to find the errors in these apps, I scroll a lot(sometimes, command + f plus app name).
now, I've this set of apps together, and looks easier to me.
---
@stevecrozz, this sort per column function looks interesting. at the first time using the app, I was looking for it.

funglaub:
Well this is rather a a matter of taste than being a real problem. I created this issue since I noticed the behaviour changed after I updated.

@lucianosousa's point makes sense too. But for me it's more imported to see what's actually having unresolved errors in a long project list :) How about making this option configurable at `/site_config`?

lucianosousa:
Too much options. I think the best approach will be some kind of table sort(per column)

stevecrozz:
Yeah maybe table column sorting is the best option. Of course as soon as we add that, the next feature will be to persist your column sorting preference. My current thinking is to restore the old behavior of sorting apps by unresolved error count. Then we can add a story about per-column sorting.

lucianosousa:
https://github.com/errbit/errbit/commit/644116d884df846f633ee0d213fab94138e80465

```
$ git revert 644116d884df846f633ee0d213fab94138e80465
```

will fix it

funglaub:
I'm fine with that, too

stevecrozz:
[...]

PR #1016, opened 2015-12-22
Ensure that App#<=> is used for sorting

md5:
Prior to this change, the `#sort` method was being handled by Mongoid and sent
to MongoDB, resulting in the `App#<=>` method being ignored for this purpose.

See additional discussion in #973

### DIFF
--- a/app/controllers/apps_controller.rb
+++ b/app/controllers/apps_controller.rb
@@ -9,7 +9,7 @@ class AppsController < ApplicationController
   expose(:app_scope) { App }
 
   expose(:apps) do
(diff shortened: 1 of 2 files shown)


md5:
This could use some better specs to ensure that the sorting behavior is actually as expected for different scenarios, but I'm pressed for time at the moment.

I may have some time in a couple weeks to add additional specs if you want to leave this PR open, otherwise please merge.

stevecrozz:
Thanks @md5. Specs would be nice. In fact I think we have some specs for this but obviously they're not good enough. I'm going to merge as-is for now and we can revisit the specs.

PR #1019, opened 2015-12-24
Break long names/errors using the word-break css3 property. Fixes #1011

### LINKED ISSUE #1011: Layout broken when Apps name is too long
It seemed that when the App name is too long, it break the layout.
break
with long name, the layout is broken.

no_break
when the name is renamed to shorter one, the layout looked fine.

I haven't tried to tinkering the CSS so I didn't know what had caused that.

cover:


stevecrozz:
From checking out the existing CSS, it seems the intent here was to use white-space: nowrap and add ellipses for long lines. Using word-break: all causes breaks to sometimes appear in weird places like p/roduction so I'm not loving the way this looks:
errbit unresolved errors

I think we should go back to white-space: nowrap and maybe even add title attributes so we can see all the text on hover.

PR #1020, opened 2015-12-25
Test against Ruby 2.3.0 on Travis CI

tricknotes:
It works fine for me :santa:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: ruby
 rvm:
   - 2.1
   - 2.2
+  - 2.3.0
   - rbx-2
   - ruby-head
   - jruby-head


stevecrozz:
Thanks! Although should we leave the .0 off the end so that we're always testing the latest 2.3 release?

tricknotes:
> we're always testing the latest 2.3 release?

This is good idea!
However Travis CI doesn't support `2.3`: https://travis-ci.org/tricknotes/errbit/jobs/99426151
To run test with Ruby 2.3, it seems `2.3.0` is required.

NOTE: Current Travis would't detect latest version of `2.2` and `2.1` :<
* `2.2` -> `2.2.2` (latest: `2.2.4`)
* `2.1` -> `2.1.5` (latest: `2.2.8`)

stevecrozz:
Interesting. I didn't realize that. I'l merge this then.

PR #1021, opened 2015-12-31
Updated configuration instructions

tradziej:
Newest version of `airbrake` notifier seems not compatible with errbit

Ref: https://github.com/airbrake/airbrake/blob/master/docs/Migration_guide_from_v4_to_v5.md


mgrachev:
:+1:

stevecrozz:
Thanks @tradziej. But I think I'd rather support v5 and it should be a pretty simple change. Let me know if you run into any issues with this approach: https://github.com/errbit/errbit/commit/ff341e5e5bb20512b7c423ce010d613dcc31fbac

PR #1025, opened 2016-01-06
Make Problem#url use the configured protocol

francois:
Configuring a protocol did not use that protocol when generating the URL, such as when used in the Slack integration.

### DIFF
--- a/spec/models/problem_spec.rb
+++ b/spec/models/problem_spec.rb
@@ -475,4 +475,26 @@ describe Problem, type: 'model' do
       end
     end
   end
+
+  context "#url" do
+    subject { Fabricate(:problem) }
+
+    it "uses the configured protocol" do
+      allow(Errbit::Config).to receive(:protocol).and_return("https")
+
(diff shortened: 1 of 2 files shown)


spec/models/problem_spec.rb
```diff
@@ -475,4 +475,28 @@
...
+      Errbit::Config.port     = 8123
+    end
+
+    after(:each) do
+      Errbit::Config.protocol = original_protocol
+      Errbit::Config.host     = original_host
+      Errbit::Config.port     = original_port
+    end
+
+    it "uses the configured protocol" do
```
stevecrozz: Does it still work if you remove all the before/after hooks and instead use this kind of thing?
```ruby
allow(Errbit::Config).to receive(:protocol).and_return('https')
```

stevecrozz:
Thanks for this @francois. I have one question that could help make it a bit cleaner and I'll be ready to merge this.

francois:
With mocks, it was as easy to use 3 different tests. This would make the tests fail differently if the URL was built incorrectly. I hope this meets your approval, @stevecrozz!

stevecrozz:
Awesome. Thanks so much!

Issue #959, opened 2015-10-06
local .env file in development can wreak havoc on specs


stevecrozz:
Any values in a local .env file affect the specs when I think they should not. In CI, there is no .env file. A developer should be able to set inputs in a local .env file without affecting the specs.

md5:
Maybe something mocking `Dotenv::Environment.new('.env')` with something like this would work: https://relishapp.com/rspec/rspec-mocks/docs/configuring-responses/calling-the-original-implementation#`and-call-original`-can-configure-a-default-response-that-can-be-overriden-for-specific-args

md5:
Per perhaps somehow mocking the `Dotenv.load` call that happens here to ignore the `.env` argument: https://github.com/bkeepers/dotenv/blob/master/lib/dotenv/rails.rb#L27-L30

stevecrozz:
I'm thinking it might be a good idea for developers to put their own configuration in a separate file, like .env.development or .env.local. This would be a gitignored file with lower precedence than .env, but if we document that it should give developers a way to set inputs that don't apply to the specs. Any thoughts?

md5:
`dotenv` loads both of those filenames by default, but the idea seems reasonable.

I'm thinking this may be worth a feature request to `dotenv`, since this is probably a more general problem.

md5:
Actually, I guess that telling developers to use `.env.development` would be fine since `.env.test` would be loaded for specs. Using `.env.local` would get loaded for all environments.

md5:
I actually tried renaming my local `.env` to `.env.development` today and I was still seeing failures with `rake spec`. I had to run it as `RAILS_ENV=test rake spec` (after renaming the file) to avoid this issue. I'm not sure why that's the case, since the `spec/spec_helper.rb` sets `ENV['RAILS_ENV'] = 'test'` before `config/environment.rb` is loaded.

stevecrozz:
@md5 What do you think about switching to the full-on dotenv-rails approach which is defined here: https://github.com/bkeepers/dotenv/blob/v2.0.2/lib/dotenv/rails.rb#L27-L31

Under this scenario I propose we check in a .env file which has sensible defaults, a .env.development file which has sensible default overrides for development and the same goes for a .env.test file. Finally we .gitignore the .env.local file which would override everything if you have it.

md5:
[...]

PR #1028, opened 2016-01-09
Refs #959 set rails env=test for test rake tasks

stevecrozz:


md5:
I don't think `default` needs to be listed. I'd probably also check `top_level_tasks.any? { |t| test_env_rake_tasks.include? (t) }`

Otherwise, this looks great.

stevecrozz:
Sweet. I made those changes and I think this is good. I'm gonna merge it.

md5:
Looks like I spoke too soon about removing `default`. I've opened #1029 to add `default` to `test_env_tasks`.

stevecrozz:
No prob

PR #1032, opened 2016-01-18
Use Ruby 2.3.0 by default for Heroku deployment

tricknotes:


felixbuenemann:
I've seen several crashes with 2.3.0 and rails and 2.3.0 ist not faster than 2.2.4 (but it has lower memory usage). SO my recommendation would be to wait for 2.3.1.

stevecrozz:
@felixbuenemann / @tricknotes Ruby 2.3.1 is the current stable version. Any opinions on making it the default for Errbit? I don't keep close tabs on what's happening with Ruby.

felixbuenemann:
I have been running some apps on 2.3.1 and it has been stable for me, so I have no objections to updating to 2.3.1.

stevecrozz:
Closing this in favor of #1130 for the 2.3.1 upgrade

Issue #1024, opened 2016-01-05
Notice for old app version ignored


StickyBros:
Hi there,

I'm trying to get airbrake-js v0.5.8 to work with errbit 0.5.0 (I also tried building the head of the master branch with the same results)

My code looks like:

```
const AirbrakeClient = require('airbrake-js');
const airbrake = new AirbrakeClient();
airbrake.setHost('http://192.168.99.100:5000');
airbrake.setProject('105cd6c8fb856...', '105cd6c8fb856...');
airbrake.notify({error: {message: 'test'}});
``` 

My app is running on localhost, so the request is going cross-domain.  I'm guessing because the response body isn't JSON it causes another error which triggers airbrake to send another request which causes an infinite loop of these:

Request

```
Accept:*/*
Accept-Encoding:gzip, deflate
Accept-Language:en-US,en;q=0.8
Cache-Control:no-cache
Connection:keep-alive
Content-Length:689
Content-Type:text/plain;charset=UTF-8
Host:192.168.99.100:5000
Origin:http://localhost:8000
Pragma:no-cache
Referer:http://localhost:8000/
User-Agent:Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_2) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/49.0.2604.0 Safari/537.36
```

What I get back is:

```
Access-Control-Allow-Headers:origin, content-type, accept
Access-Control-Allow-Origin:*
Cache-Control:max-age=0, private, must-revalidate
Connection:close
Content-Encoding:gzip
Content-Type:text/html; charset=utf-8
Date:Tue, 05 Jan 2016 21:04:41 GMT
ETag:W/"d75e9c79cd8a749c7fc2094f74d81a36"
Status:200 OK
Vary:Accept-Encoding
X-Content-Type-Options:nosniff
X-Frame-Options:SAMEORIGIN
X-Request-Id:ac075876-6574-4c69-a529-37fc323ca577
X-Runtime:0.004290
X-XSS-Protection:1; mode=block
```

with the body

```
Notice for old app version ignored
```

stevecrozz:
It looks like we have some code that's intending to reject error notifications for old versions. Can you check to see what you have set for LATEST APP VERSION for this Errbit app? If I understand the intent correctly, you'll need to pass a newer app version in the notification than the one you have set in your Errbit app.

StickyBros:
Ah, good catch.

[...]

PR #1041, opened 2016-02-04
Airbrake 5.0 expects 201 status code for v3 api

felixbuenemann:
When using the airbrake-ruby gem (used by airbrake 5.0), the client throws an error due to a bad return code from the v3 api controller.

The expected return code is 201 created, but currently the errbit controller returns 200 ok.

See: https://github.com/airbrake/airbrake-ruby/blob/master/lib/airbrake-ruby/response.rb#L22

### DIFF
--- a/app/controllers/api/v3/notices_controller.rb
+++ b/app/controllers/api/v3/notices_controller.rb
@@ -18,7 +18,7 @@ class Api::V3::NoticesController < ApplicationController
     return render text: VERSION_TOO_OLD, status: 422 unless report.should_keep?
 
     report.generate_notice!
-    render status: 200, json: {
+    render status: 201, json: {
       id:  report.notice.id,
       url: report.problem.url
     }
--- a/spec/controllers/api/v3/notices_controller_spec.rb
+++ b/spec/controllers/api/v3/notices_controller_spec.rb
@@ -21,9 +21,9 @@ describe Api::V3::NoticesController, type: :controller do
     )
   end
 
-  it 'responds with 200 created on success' do
+  it 'responds with 201 created on success' do
     post :create, legit_body, legit_params
-    expect(response.status).to be(200)
+    expect(response.status).to be(201)
   end
 
   it 'responds with 400 when request attributes are not valid' do


stevecrozz:
Yep. This may break certain older clients, but we don't have the bandwidth to support everything, so I'm good with this.

felixbuenemann:
Hmm, that would be strange, then the official airbrake api would have to do user agent sniffing. 

The v1 api should not be affected.

Do you know which clients expect the 200 code?

stevecrozz:
I don't know offhand, but I'm sure I've seen it. It may have been an older version of the airbrake-js client...

kimrgrey:
@stevecrozz, take a look on this comment: https://github.com/errbit/errbit/issues/1024#issuecomment-169166551. May be it will clarify a problem.

stevecrozz:
ugh. yeah that looks familiar. I honestly don't know what to do about this. One client wants the response to be 200, the other wants 201. Both clients are officially supported by airbrake...

I really don't want to add user agent sniffing stuff or change the API responses for different clients.

felixbuenemann:
The current version of airbrake-js has an "xhr" reporter that expects 201 and a "compat" reporter that expects 200 as well as a "jsonp" and "console" reporter. From a quick read it seems that the default is "compat" if credentials are used and "jsonp" otherwise.

Maybe we should dig into this before doing a new release.

stevecrozz:
@felixbuenemann I think what you're looking at is https://github.com/airbrake/airbrake-js/blob/v0.5.8/src/client.coffee#L40 I think that's essentially feature detection with a jsonp fallback. Modern browsers should get the compat reporter always.

felixbuenemann:
Interestingly it seems airbrake-js between v0.4.3 and v0.5.3 expected status 200. Older than v0.4.3 expect 201 and newer than v0.5.3 expect 200 with the compat reporter and 201 with the xhr reporter.

kimrgrey:
May be it's good idea to create an issue in airbrake-js, because official documentation is definitely talking about 201 Created... :smirk:

felixbuenemann:
@stevecrozz It doesn't look like airbrake supports jsonp, so the client would have to be configured manually anyways. In which case one can just set the reporter to "xhr".

I'm not sure how to detect older clients, so probably we could just tell people to update their client.

felixbuenemann:
@kimgrey Yeah, that's a good idea. But there's still the question of how to handle older clients in the wild.

kimrgrey:
@felixbuenemann, from the errbit's API side? Well, as for me, if it's just a bug in this library, I don't think that it is a right way to support wrong expectations. We can just add some notice in the changelog for the new release.

felixbuenemann:
@stevecrozz How about adding a configuration setting for this? Seems less painful than doing client detection. This way we could use 201 for new installations and have old installations continue to use 200 unless their config is updated manually.

stevecrozz:
In the past, I've been pretty aggressive on insisting people upgrade their client libraries and that's fine with me. If you really want to support older clients, you can use an older Errbit :)

But there's an extra wrinkle this time because the latest version of the official JavaScript client will not work by default. I've raised an issue about it here: https://github.com/airbrake/airbrake-js/issues/189

felixbuenemann:
At least of the airbrake-js client, switching to 201 should be harmless. If would only cause clients to not use `console.debug` to log the error id and url.

felixbuenemann:
Oh, it seems in current versions there's a callback attached.

kimrgrey:
>  If you really want to support older clients, you can use an older Errbit :)

That's true :+1: :smile:

felixbuenemann:
@stevecrozz I'm fine with not supporting older clients, but it seems the airbrake-js example in the README needs updating and we could mention the proper setup and required version there.

stevecrozz:
Yeah. Let's see if we can get an answer on that airbrake-js issue I opened in the next day or so. If not we can find a workaround.

felixbuenemann:
From reading the source and README of airbrake-js the config should look like:

```javascript
var airbrake = new airbrakeJs.Client({
  projectKey: 'ERRBIT API KEY',
  reporter: 'xhr',
  host: 'http://errbit.yourdomain.com'
});
```

stevecrozz:
I got my answer from airbrake/airbrake-js#189 and it looks like it confirms your recommendation @felixbuenemann. I can make the documentation change and that should settle things.

stevecrozz:
Looks like there's a bit more work to do to make errbit compatible with the latest airbrake-js. I'm opening an issue about this: https://github.com/errbit/errbit/issues/1044

PR #1043, opened 2016-02-04
Backtrace decorator should not raise an error when no lines present.

kimrgrey:
`BacktraceDecorator` raises "undefined method `map' for nil:NilClass" when backtrace contains no lines. After upgrading to latest Airbrake version we receive this error very often. So, I think it's good idea to handle this situation.

screen shot 2016-02-04 at 15 13 46

### DIFF
--- a/app/decorators/backtrace_decorator.rb
+++ b/app/decorators/backtrace_decorator.rb
@@ -1,6 +1,6 @@
 class BacktraceDecorator < Draper::Decorator
   def lines
-    @lines ||= object.lines.map { |line| BacktraceLineDecorator.new line }
+    @lines ||= Array(object.lines).map { |line| BacktraceLineDecorator.new line }
   end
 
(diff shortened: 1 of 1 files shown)


stevecrozz:
Seems like a good thing to protect against, although I'm inclined to think we should be more defensive and validate backtrace lines be not nil instead. That, plus adding a spec to make sure it doesn't regress in the future would be great. Are you down for that @kimrgrey?

kimrgrey:
Do you prefer to add a validation in the model instead to not allow nil value at all? Should we replace nil with an empty array, for example, in before_validation callback to be sure that we always have something?

stevecrozz:
Yeah I think that sounds good. Do you agree?

kimrgrey:
Sure =) So, I gonna close this PR and send another one when things will be done :wink:

filipkis:
Any news on this?

PR #1046, opened 2016-02-19
Dokku deployment support

ebeigarts:
I have successfully deployed errbit to dokku and [here's the output](https://gist.github.com/ebeigarts/8c3687eddd62b339fe05).

There was an [issue](https://github.com/dokku/dokku/issues/1899) with the min ruby check, so I moved it to config/boot.rb.

### DIFF
--- a/CHECKS
+++ b/CHECKS
@@ -1,0 +1,1 @@
+/users/sign_in Errbit
--- a/Gemfile
+++ b/Gemfile
@@ -4,14 +4,6 @@ RAILS_VERSION = '~> 4.2.5.1'
 
 send :ruby, ENV['GEMFILE_RUBY_VERSION'] if ENV['GEMFILE_RUBY_VERSION']
 
-detected_ruby_version = Gem::Version.new(RUBY_VERSION.dup)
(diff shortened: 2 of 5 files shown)


stevecrozz:
This is very cool. Thanks for taking the time to put it together. I have zero familiarity with dokku other than reading about it once, so I do have a few questions:

1. Why do we need this CHECKS file? Does it have to be in the project root?
2. Can you explain the issue with the minimum ruby version check?

Lastly, I have to admit that I don't have time to test and maintain new deployment scenarios. If you can volunteer to answer other people's questions and fix bugs with this implementation/docs, then we should be good to go with the merge after talking about the questions above.

mattberther:
Although not the original author of the pull request, I believe this is a valuable addition -- and just like @ebeigarts, Im planning to stick with errbit and dokku so Id be happy to help out also.

stevecrozz:
I am surprised dokku uses ruby 1.9.3 at all since that was end-of-lifed about this time last year. But it sounds like you've got a good workaround and already this was already seconded by @mattberther so I'll merge it. Thanks for the contribution.

PR #1056, opened 2016-02-20
closes #1053 don't carry forward flash notices

### LINKED ISSUE #1053: app save failure notice carrying over to next page
When the application update fails, we get a message stating so, as shown here
image
and then, when I click on another page, like the users, the error message carries over...
image
probably needs the controller to set the message using `flash.now` instead of just `flash`

mfrederickson:
Uses flash.now instead of just flash when rendering and not redirecting.


stevecrozz:
Looks good. Thanks.

PR #1058, opened 2016-02-20
Fix 1049 username testing

### LINKED ISSUE #1049: many tests fail when ERRBIT_USER_HAS_USERNAME=true
I had this set in my dev environment and ran `rake spec` and had a couple hundred failures... almost gave me a heart attack ;-) Then I changed it back to false and reran the tests and only had a couple failures (of my own making).  Is this something that should be avoided, or fixed?

mfrederickson:
Changed to set username during user fabrication, and use that field for authentication, if username config set.  Closes #1049.

### DIFF
--- a/spec/acceptance/acceptance_helper.rb
+++ b/spec/acceptance/acceptance_helper.rb
@@ -22,7 +22,13 @@ end
 
 def log_in(user)
   visit '/'
-  fill_in :user_email, with: user.email
+
+  if Errbit::Config.user_has_username
+    fill_in :user_username, with: user.username
+  else
+    fill_in :user_email, with: user.email
+  end
+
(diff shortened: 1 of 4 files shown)


stevecrozz:
These tests will be testing different things depending on the provided configuration. I think the tests should be explicit in what they are testing. Instead of checking to see if `Errbit::Config.user_has_username` is truthy, I would prefer to `allow(Errbit::Config).to receive(:user_has_username).and_return(true)` or `allow(Errbit::Config).to receive(:user_has_username).and_return(false)`

stevecrozz:
There's a few examples of this hanging around. Here is one:
https://github.com/errbit/errbit/blob/v0.6.1/spec/views/problems/show.html.haml_spec.rb#L78

PR #1063, opened 2016-02-22
Bootstrap ui

mfrederickson:
I needed something a little more responsive, so this is an attempt to "bootstrapify" the existing ui.  Although I do like some of the other proposed bootstrap layouts better, this meets my needs.  Let me know what you think.


stevecrozz:
Impressive. I'm going to prioritize having a look at this one.

mfrederickson:
(Sorry about the three squashes, the first two didn't work out like I'd expected.)

stevecrozz:
I'm looking at the apps index which is the first thing I'm noticing. This was formerly a table with a bit more information about apps including repository name which seems useful. Do you have a reason for making this change? or is this part still a work in progress?
screenshot-localhost 9292 2016-03-01 08-50-55

stevecrozz:
Yeah, that's it. What went into the decision to ditch the table? At first glance, I'd be inclined to keep the table and simplify the column display by always displaying all the columns. But I'd be curious to see if you had some reasons for doing it differently.

mfrederickson:
Tables aren't responsive -- they don't react to different screen sizes well.  To be able to show all the information nicely on a narrow screen the row has to actually have the cells stack.

stevecrozz:
[nimbus-record-video.webm.zip](https://github.com/errbit/errbit/files/157210/nimbus-record-video.webm.zip)

It's true that using the css grid makes the tabular data responsive, but there's a couple things that I'm not convinced about yet. The way the column headings stack (name, repository) seems a bit odd since they're not able to describe a 'column' in the same way. Tables have another nice property that we'll miss out on where they automatically attempt to scale the columns to fit whatever content is inside. Not a huge deal if we decide to abandon tables here but it is kind of a nice feature if we go with tables. What do you think about http://getbootstrap.com/css/#tables-responsive ?

I think I also did enjoy having the error count in its own column for scannability and for the potential of making it a sortable column down the road.

Gemfile
```diff
@@ -99,11 +99,13 @@ gem 'puma'
...
 gem 'uglifier'
-# We can't upgrade because not compatible to jquery >= 1.9.
-# To do that, we need fix the rails.js
-gem 'jquery-rails', '~> 2.1.4'
+gem 'jquery-rails', '~> 4.1'
 gem 'pjax_rails'
 gem 'underscore-rails'
 
+gem 'bootstrap-sass', '~> 3.3.6'
+gem 'coffee-rails'
```
thejspr: There doesn't appear to be any coffeescript in the project, do we need this gem then? (Awesome work btw, I really hope this gets merged! :+1: )

mfrederickson:
@thejspr Thanks.  I had trouble with the twitter-bootstrap-rails-confirm until I added it.
image
And when I removed it just now, I get
image

mfrederickson:
@stevecrozz I've not looked at the zip yet, but in regards to the tables-responsive, I'm actually using that in the notice details for some of the tabs because that was quick and I thought I'd come back to it later (and I still might).  But why have a table to scroll horizontally through when you're not comparing columns?  And I always thought that if you prefer the table looking layout (still provided by the responsive grid) then use a larger screen device, but if you had to use a smaller screen device, it would still be usable and stacking in that case made sense to me, but if you'd prefer a different responsive layout, I can work on that (time permitting) just need a little direction...

mfrederickson:
Ahh, the dreaded "W"  (re: screen recording).  Well, you can have CSS word-breaking place nice (at the spaces):
image

Or not so nicely (whereever apparently "needed"):  `word-break: all`
image

Though it does make the superlong W application respect it's boundaries.  I'll commit that change.  Perhaps someone with a little more CSS skill than I can tweak it to be perfect?

thejspr:
@mfrederickson there is an option that doesn't require coffeescript https://github.com/ifad/data-confirm-modal

Maybe even using the standard browser alert box is ok? OR maybe we don't even need the confirm box? Possibly a discussion for another time though.

stevecrozz:
I would prefer the defautl javascript confirm boxes to bringing in coffeescript, I'm also still thinking about the tables. I think you can get it looking pretty good without tables, but here's a quick screenshot of some funkiness I've observed:
screenshot-localhost 9292 2016-03-22 22-52-03

I think to make these 'columns' work well, they'll need to have a minimum size and all the content will need text-overflow: ellipses with title attributes so the full content is visible on hover. It can be pretty hard to use if you hit the wrong screen size and the content isn't as expected.

Another thing is that fieldset and legend elements look bad in plain bootstrap, so that makes the forms that use them look pretty bad here. What do you think about using bootstrap panels instead of our old fieldsets?

mfrederickson:
Could you explain the hesitation regarding the Coffeescript? I agree that
slimmer is better, but isn't this processed into javascript and compressed
into the asset pipeline?

stevecrozz:
Sure. I don't want to open up too large a can of worms on coffeescript, but essentially I see it as an unnecessary dependency. One extra thing to keep up-to-date and an extra annoyance every time I use a rails generator and I accidentally generate coffee files instead of regular js files. I also plain dislike coffeescript but this wouldn't be an appropriate place for me to vent about coffeescript :)

thejspr:
I don't see coffeescript as a blocker, would rather get this into master :) I've been running this branch on my production errbit for a while and not noticed any bugs so far.

stevecrozz:
I agree about coffeescript. One thing that's holding me back is that the look and feel (design elements) of the new bootstrapified system are a step backward. My guess is that we'll get a lot of complaints about that if we merge as-is. See my last screenshot for an example of some of this ugliness where 'about two months ago' wraps into 'abo<break>ut 2<break>mon<break>ths<break>ago'

Now, we don't need to block this PR on making the design perfect as that'll mean we never merge. But I think there's some things that need doing that I tried to call out. We could find a time to hash out a review in real time and build a list of blocking issues if you have any interest in helping push this along. Or I could get in here and do some serious QA and log all the issues I find. I think if you start from a brand new Errbit install and go through the UI with varying browser resolutions, you'll find plenty of issues.

Alternatively, we could still break this up into smaller, more manageable PRs.

mfrederickson:
Sorry I've taken a back seat on this... my priorities have shifted a little, especially since this is functioning well for us.  Though I do agree about some tweaks that would make it a little better.  I know the CSS change to accommodate a really long non-breaking project name impacts the breakpoints for wrapping other items such as the "time ago".  Targeting that CSS to specific columns would help some.

rud:
It seems master has veered from this PR and reviving it will take a bit of heroic effort. Any takers?

PR #1064, opened 2016-02-22
incorrect namespace for rake task in docs

stephancom:


stevecrozz:
Nice. Thanks for the catch. Although this incorrect text seems to appear elsewhere. Would you like to change the other references as well?

stephancom:
I'll try to get to it soon :)

stephancom:
appears to be done - was also referenced in another rake task

stevecrozz:
Thanks @stephancom

PR #1065, opened 2016-02-22
closes #1051 problems table layout

### LINKED ISSUE #1051: minor problem with problem table view
It seems to possibly be short a colspan or something...
image
... not to mention that fact that the table is extending beyond it's "container".

mfrederickson:
fixes #1051
image


born4new:
Why not deleting the lines altogether, instead of leaving commented-out code?

danielwillemse:
I think this fix introduces a new problem with the ui.
screen shot 2016-11-03 at 09 57 22

rud:
Most likely fixed by  #1217

PR #1066, opened 2016-02-23
fix searching when search field is empty

### LINKED ISSUE #1054: text search/sort problem
When I search for problems and the list comes back empty, and then I click on a column header to sort, it blows up...  (doesn't make sense to do this, but still something that should be fixed)...
image

mfrederickson:
closes #1054 by skipping the search part of the scope when no search criteria is specified and added a test example to ensure searching without criteria returned a result set instead of blowing up.

### DIFF
--- a/app/controllers/problems_controller.rb
+++ b/app/controllers/problems_controller.rb
@@ -124,7 +124,11 @@ class ProblemsController < ApplicationController
   end
 
   def search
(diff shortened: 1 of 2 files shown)


mfrederickson:
(The travis failure is a monodb unreachable error again on the last test...)

stevecrozz:
@mfrederickson I plucked your commit onto a new up-to-date branch, keeping the test and the heart of the fix, but adding a minor refactor. The PR is ready on #1131.

PR #1067, opened 2016-02-23
preserving sorting when showing all/unresovled and searching

### LINKED ISSUE #1052: sorting lost when show un/resolved button clicked
I would expect the sorting to remain in place while on that page... but if you sort by a column such as the 'What and Where' and then choose to see resolved errors, the sorting is lost.
image

mfrederickson:
closes #1052. also retains sorting when searching.


stevecrozz:
Thanks @mfrederickson. I let the branch get too stale and had to manually merge to resolve a conflict. But your fix is now in master.

PR #1068, opened 2016-02-26
puma updated to version 3

otzy007:
Hi,

I've updated puma to v3 :smile:


stevecrozz:
seems like a good idea

PR #1075, opened 2016-03-18
Get hostname from the hostname key

aminin:
According to the [API V3 docs](http://airbrake.io/docs/#create-notice-v3) context has a separate key for _hostname_.

When I have 2 app servers: `app01.infra.example.com` and `app02.infra.example.com`  those handle similar requests, I want to realize which one produces the error.

In this case parsing the URL doesn't make sense.

### DIFF
--- a/lib/airbrake_api/v3/notice_parser.rb
+++ b/lib/airbrake_api/v3/notice_parser.rb
@@ -84,7 +84,7 @@ module AirbrakeApi
       end
 
       def hostname
-        URI.parse(url).hostname
+        context['hostname'] || URI.parse(url).hostname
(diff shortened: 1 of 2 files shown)


jasherai:
:+1:

stevecrozz:
@aminin Seems like a good change. Would you like to introduce a test case to make sure we don't mistakenly break this functionality down the road?

aminin:
@stevecrozz I've introduced the test case, and have no idea why coverage decreased.

trv1:
Yeah, baby, I waited for this for ages.

stevecrozz:
Looks good. Thanks!

PR #1076, opened 2016-03-23
Add icon for slack

ebeigarts:
screen shot 2016-03-23 at 10 35 59


spec/models/notification_service/slack_service_spec.rb
```diff
@@ -9,14 +9,18 @@
...
+
+  it "should send a notification to Slack with hook url" do
     # setup
     problem = notice.problem
 
     # faraday stubbing
     payload = {
       username:    "Errbit",
-      icon_emoji:  ":collision:",
+      icon_url:    "https://raw.githubusercontent.com/errbit/errbit/master/docs/notifications/slack/errbit.png",
```
stevecrozz: Is this normal? It seems a little strange to load the icon through raw.githubusercontent.com. Wouldn't it be more normal to put the icon in the errbit application and let errbit host it?

ebeigarts: I guess the only argument is that errbit could be installed behind a firewall etc, and in that case it wouldn't work.

stevecrozz: Fair enough. I guess if github doesn't mind us hotlinking like this then its fine with me.

PR #1078, opened 2016-04-07
Add omniauth by google

8398a7:
I was implemented authentication by google to errbit.

demo

The reason for this is that, GitHub is offen unusable.
In the authentication by Google, such a situation is considered unlikely to occur.

Implementation is the following three points.
- Authentication by google
- Test of authentication by google
- Update of the document

Please tell me if there is a correction point:dog:


stevecrozz:
Seems like a good idea to me.

PR #1079, opened 2016-04-08
Add gem 'capistrano-rvm' as option to rbenv

adisos:
It's useful for those who are using rvm instead of rbenv.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -69,6 +69,7 @@ group :development do
   gem 'capistrano-rails',   require: false
   gem 'capistrano-rbenv',   require: false
   gem 'capistrano3-puma',   require: false
+  gem 'capistrano-rvm',     require: false, github: 'capistrano/rvm'
 
   # better errors
   gem 'better_errors'
--- a/Gemfile
+++ b/Gemfile
@@ -69,7 +69,7 @@ group :development do
   gem 'capistrano-rails',   require: false
   gem 'capistrano-rbenv',   require: false
   gem 'capistrano3-puma',   require: false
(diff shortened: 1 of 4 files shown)


Gemfile
```diff
@@ -69,6 +69,7 @@ group :development do
   gem 'capistrano-rails',   require: false
   gem 'capistrano-rbenv',   require: false
   gem 'capistrano3-puma',   require: false
+  gem 'capistrano-rvm',     require: false, github: 'capistrano/rvm'
```
mfrederickson: I'm curious why you are specifying the github repo.

stevecrozz:
Is rbenv not working for you @adisos?

Gemfile
adisos: Removed that line

adisos:
@stevecrozz 

I think it's not that bad to have more options to deploy. I don't like rbenv, I'm using rvm all the time, so I think I'm not only one who's also using it.

stevecrozz:
@adisos That makes sense. And I don't mind adding a reference to rvm in the deploy.example.rb file, but I'm not enthusiastic about adding a new Gemfile dependency for everyone.

Would you be ok with adding to the deploy.example.rb file something more like:
"For rvm, add capistrano-rvm to UserGemfile and see https://github.com/capistrano/rvm#usage"

We could always treat rbenv the same way if we feel the need.

stevecrozz:
Rvm is a great tool, but I'm closing this because I don't think Errbit needs more options in this area. For others curious about this, I'd be more interested in advancing the container-based deployment workflow where all of these concerns are hidden away inside the container so that (hopefully) nobody has to worry about them.

PR #1082, opened 2016-04-14
automatically create issue

quexpl:
I've created two changes in errbit. First one is adding sidekiq jobs for notifications - there is no need to wait until notifications are sent, it slows down errbit client. 
Second is ability to automatically create issue when first error occurs. This option (like other) can be set globally or per app.


stevecrozz:
Thanks for bringing this up @quexpl. First thought is I think it would be a good idea to split these conversations. Let's talk here about automatically creating issues. The idea makes sense to me. I think I'd like to see a bit of support from the community as every feature we add increases the complexity of Errbit. But if we can get buy-in from a few different Errbit users, or even a few +1s here, then we can move forward with the process which would involve adding docs and specs and getting this part merged.

As for the async notifications, this has come up before so it may be a good idea for you to read through the last conversation we had about it here: https://github.com/errbit/errbit/pull/834. I think there is some appetite for this. If you want to pursue it would you open a new issue where we can talk just about async notifications?

Strnadj:
@quexpl We discuss about that with @stevecrozz, but unfortunattely I dont have a time to finish it yet... This is not a good way actually (i've already proposed that in https://github.com/errbit/errbit/pull/834/files), but why forcing user to use sidekiq... you can do that better https://github.com/Strnadj/errbit/commit/4f389a5b47f42483a5cb38fc78e41a3e1c3df98f, let user to add their own "favourite" background processing manager into UserGemfile.

Strnadj:
You can steal my idea with ActiveJob with :inline adapter :) and finish it.. no problem :+1:

quexpl:
@Strnadj this is great idea! I'm moving discussion about background jobs to issue #1084 and here leave place to discussion about automatically create issue.

rud:
Closing due to inactivity, hope that's okay.

PR #1090, opened 2016-04-21
Add MongoDB key for heroku

trueinviso:
Default heroku key is MONGODB_URI which wasn't included in configuration.


stevecrozz:
It looks like @mattberther beat you to it by a hair, but thanks anyway.

PR #1092, opened 2016-04-26
First setup the config files and later check

otzy007:
Please correct me if I'm wrong but first the config files should be uploaded to the destination server and later check the configs.

If I run first `bundle exec cap production deploy:check` I'll get this error:
```
ERROR linked file /var/www/apps/errbit/shared/config/puma.rb does not exist on 
(Backtrace restricted to imported tasks)
```
Running first `bundle exec cap production errbit:setup` will pass


stevecrozz:
You're right. This is probably a better order. I think the intention here was to show how a failed check works, but maybe that doesn't make a lot of sense.

PR #1095, opened 2016-05-10
Slack rooms

jaxxstorm:
currently, the slack notifier only allows you to post to the channel the webhook is configured with by default. This PR aligns the notifier with others in that it asks you pick the channel to send notifications to by default

### DIFF
--- a/app/models/notification_services/slack_service.rb
+++ b/app/models/notification_services/slack_service.rb
@@ -4,12 +4,16 @@ class NotificationServices::SlackService < NotificationService
     [:service_url, {
       placeholder: 'Slack Hook URL (https://hooks.slack.com/services/XXXXXXXXX/XXXXXXXXX/XXXXXXXXX)',
       label:       'Hook URL'
+    }],
+    [:room_id, {
+      placeholder: '#general',
+      label:       'Room where slack should notify'
     }]
   ]
 
   def check_params
     if FIELDS.detect { |f| self[f[0]].blank? }
-      errors.add :base, "You must specify your Slack Hook url."
+      errors.add :base, "You must specify your Slack Hook url and channel"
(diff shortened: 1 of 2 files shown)


stevecrozz:
This is good. The only thing I want to discuss it to make sure users who set up these notifications prior to this change will not experience any adverse side effects. These users would not have specified a channel when they set up slack notifications and so if they grab this new code and visit the app edit page, I think they may not be able to save the app because the notifier has a new validation that says channel must be present.

Is that right, and if so do you have any suggestions?

jaxxstorm:
This is correct. What I _could_ do is modify the `FIELDS.detect` check to ignore the room_id, and then make this optional with a method that checks if that FIELD exists, and uses the default if not. Thoughts?

stevecrozz:
I think that sounds good.

stevecrozz:
@jaxxstorm Are you planning to make that change still?

jaxxstorm:
yes, I just haven't got round to it yet, sorry!

gizotti:
Any update on this one? Would come in handy for my team as well.

I could pick up from where you left if you don't hav ethe time @jaxxstorm 😃

jaxxstorm:
@gizotti we're no longer using errbit, so feel free to take this on

gizotti:
@jaxxstorm will do, thanks.

@stevecrozz what's the process you guys would prefer from here? Close this PR and I start a new one?

stevecrozz:
@gizotti that's probably the easiest way. I can close this if you make a new PR.

gizotti:
@stevecrozz no worries. Will do it tonight when I get home.

stevecrozz:
Closing in favor of #1145

PR #1097, opened 2016-05-12
Update LDAP documentation

randaalex:
According to [configuration page](https://github.com/errbit/errbit/blob/master/docs/configuration.md) params has name with `ERRBIT_` prefix


stevecrozz:
Thanks @randaalex

Issue #1102, opened 2016-06-07
`project_id = true` does not work with airbrake 5.4


delphaber:
`airbrake-ruby` 1.4.1 does a `.to_i` on `project_id` value, so we cannot use `true` anymore.

Can we use a fake number like `99999` ?

delphaber:
```bash
↳ $ rake airbrake:test
/Users/faber/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/airbrake-ruby-1.4.1/lib/airbrake-ruby/config.rb:133:in `valid?': undefined method `to_i' for true:TrueClass (NoMethodError)

````

nofxx:
Confirm, also: that's from `airbrake-ruby`, not the same gem. The version works: `1.3.1`, fails: `1.4.1`
Related to #1087

weh:
@stevecrozz issue can be closed (?)

stevecrozz:
Yep. You're right @weh

PR #1103, opened 2016-06-07
fix does not work with airbrake 5.4 refs #1102

8398a7:
refs #1102 
fix docs for airbrake 5.4

https://github.com/airbrake/airbrake-ruby/blob/989187744e7ea65f3383aacbac0e0f27500ce62a/lib/airbrake-ruby/config.rb#L133

In 5.4, an exception occurs at the point of the case `to_i` project_id is true.
project_id was work in -1, but it is first aid.


stevecrozz:
1195423550187356949molumen_red_approved_stamp svg hi

PR #1108, opened 2016-06-25
Docker and Kubernetes 

pigri:
I create Docker and Kubernetes support for Errbit. Small modifications for Gemfile because alpine linux  is have bug. ( http://forum.alpinelinux.org/forum/installation/install-therubyracer-libv8 ).


stevecrozz:
Awesome! I'm checking it all out now. I do want to point out one potential problem and that is that for whatever reason @ddollar recommends in his README file not to install foreman in a project's Gemfile. I never figured out exactly what that would be bad, but since we only run a single process, it doesn't gain us much. Would you be willing to change from: `CMD ["/usr/bin/foreman","start","web","-d","app"]` to `CMD ["bundle","exec","puma"]`?

stevecrozz:
Also, do you know why bigdecimal is not found for some of the builds? https://travis-ci.org/errbit/errbit/jobs/140259530

pigri:
Can you rerun the build please? :) I don't no why failed. My travis build it's works: https://travis-ci.org/ghostmonitor/errbit. I fixed the bigdecimal gem problem removed for Gemfile because is fixed from Dockerfile. The foreman problem...my solution is install in Dockerfile and no have problem. So appropriate for you?

stevecrozz:
Yeah this is fine. I'm not sure why foreman is needed, but I suppose it has the benefit of sharing the Procfile with heroku deployments. I'm rerunning the build now.

ddollar:
As some of the other commenters have mentioned Foreman probably isn't doing much for you here. If you want a lightweight way to run a Procfile I'd suggest checking out https://github.com/ddollar/forego

To answer the question posed by @stevecrozz I recommend that Foreman stay out of your Gemfile because we had several releases where bad interactions between the dependencies of Rails and Foreman caused problems for a subset of users. It's much easier to treat Foreman like a system tool that just happens to be built in Ruby instead of part of the resolved dependency tree of your application. 

(You can just `gem install foreman` if you need to) and avoid the potential problems but I recommend downloading the single-binary forego if you're doing this in Docker.

stevecrozz:
Thanks for the answers and suggestions @ddollar.

The build is now green. I'll do some playing around with this PR since I'm very new to kubernetes.

pigri:
Thanks for the answers and suggestions @ddollar. I change the Dockerfile your instructions.
@stevecrozz Default server puma in the Dockerfile but have alternative Dockerfile.forego.

stevecrozz:
docker build works
docker-compose up works

This mostly supercedes the work I did last year as part of https://github.com/errbit/errbit/pull/822 which didn't get a whole lot of support. There are some useful docs from that that should probably be reworked and added to your effort after this merge. I think its a good time to bring docker support in as a supported deployment method for Errbit so thanks for opening this PR.

I'm not sure yet about kubernetes. I may add a note about kubernetes having "experimental" support or something along those lines, mainly because I barely understand it and I'm not comfortable claiming that I can support it.

PR #1117, opened 2016-08-29
Fix UrlGenerationError exception thrown when Google auth is disabled

vevix:
If you disable Google authentication, the `/users/sign_in` page throws an exception of type: `ActionController::UrlGenerationError`


stevecrozz:
Thanks

PR #1118, opened 2016-08-30
more I18n and pt-BR translations

rcnetto:
The current internationalization keys weren't enough to allow a full translation of Errbit. I added all the missing keys and their corresponding pt-BR translation. The rails.pt-BR.yml was introduced to translate "time in words" functions. All tests passed with success.

### DIFF
--- a/app/controllers/problems_controller.rb
+++ b/app/controllers/problems_controller.rb
@@ -77,7 +77,7 @@ class ProblemsController < ApplicationController
 
   def resolve
     problem.resolve!
-    flash[:success] = 'Great news everyone! The error has been resolved.'
+    flash[:success] = t('.the_error_has_been_resolved')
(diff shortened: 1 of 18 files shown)


stevecrozz:
Very awesome. I'll give it a try!

stevecrozz:
I had to add to my config/application.rb.
```ruby
config.i18n.available_locales = ["pt-BR"]
config.i18n.default_locale = :'pt-BR'
```

Is there an easier way? should we document this?

rcnetto:
I think it's a good idea documenting it. It was not obvious to me (non ruby developer) how to do that. : )

stevecrozz:
I'll put that on my list for documentation, but I don't see that as a reason to keep from merging this. Thanks very much!

PR #1121, opened 2016-09-28
Remind heroku button users to bootstrap

wesgarrison:
I ran through the Heroku installer and didn't have an admin user. I had to bootstrap to set it up.

This is lower in the page, but it wasn't clear where the manual instructions ended.  We could say "finish configuration (link)" instead of this if you think that's better.


stevecrozz:
I'll add this temporarily, but I think what I should next is use https://devcenter.heroku.com/articles/app-json-schema#scripts to specify that heroku should run this script automatically. I'm thinking it would be a good idea to combine this with environment variables that define the admin user and password so users don't have to go hunting for them. I'll open a new ticket for this.

PR #1124, opened 2016-10-07
add phantomjs gem to fix failing acceptance tests

weh:
When i run rspec spec the acceptance tests are failing with the following error message:

_Cliver::Dependency::NotFound:
  Could not find an executable ["phantomjs"] on your path._

I had to add phantomjs to the Gemfile [as described here](https://github.com/colszowka/phantomjs-gem#usage-with-poltergeistcapybara)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -88,6 +88,7 @@ group :test do
   gem 'fabrication'
   gem 'capybara'
   gem 'poltergeist'
+  gem 'phantomjs', require: 'phantomjs/poltergeist'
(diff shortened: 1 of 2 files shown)


weh:
@stevecrozz Looks like phantomjs is causing problems with travis #1125 🤔

stevecrozz:
@weh Seems like some kind of temporary issue with the phantomjs archive. I've restarted the failed builds and they seem to be doing better.

PR #1126, opened 2016-10-08
Add UI for per-app fingerprinting

crunis:
Work in progress. Still needs: 

 * Get rid of the site-wide notification fingerprinting UI or add a note in that screen that is going to override any per-app fingerprinting that is set up.
 * Add the source field to `notice_fingerprinter.attributes`

### DIFF
--- a/app/views/apps/_fields.html.haml
+++ b/app/views/apps/_fields.html.haml
@@ -37,6 +37,35 @@
       = f.text_field :email_at_notices, :value => app_decorate.email_at_notices
       .field-helpertext times an error occurs (comma separated).
 
+%fieldset
+  %legend= t('.notice_fingerprinter')
+  %p= t('.notice_fingerprinter_text')
+
+  = f.fields_for :notice_fingerprinter do |g|
+    .checkbox
+      = g.check_box :error_class
+      = g.label :error_class, t('.error_class')
+
+    .checkbox
+      = g.check_box :message
+      = g.label :message, t('.message')
+
+    %div
+      = g.label :backtrace_lines, t('.backtrace_lines')
+      = g.text_field :backtrace_lines
+
+    .checkbox
+      = g.check_box :component
+      = g.label :component, t('.component')
+
+    .checkbox
+      = g.check_box :action
+      = g.label :action, t('.action')
+
+    .checkbox
+      = g.check_box :environment_name
+      = g.label :environment_name, t('.environment_name')
+
 %div.checkbox
   = f.check_box :notify_all_users
   = f.label :notify_all_users, t('.notify_all_users')


crunis:
Added a checkbox to select the usage of site wide fingerprinter or custom app one.
The attribute "source" of the noticefingerprinter is now correctly updated

app/assets/javascripts/apps.edit.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * Created by chteijon on 09/10/2016.
+ */
+
+
+$(function() {
+    $('#other_use_site_fingerprinter').change(function() {
```
stevecrozz: Lines 8-12 could be replaced with:
```javascript
$('.custom_notice_fingerprinter').toggle(!this.checked);
```

app/decorators/app_decorator.rb
```diff
@@ -7,6 +7,18 @@ def email_at_notices
     object.email_at_notices.join(', ')
   end
 
+  def use_site_fingerprinter
```
stevecrozz: I find it easier to read when I break these down onto multiple lines of guard clauses:
```ruby
def use_site_fingerprinter
  return true if object.notice_fingerprinter.nil?
  return true if object.notice_fingerprinter.attributes['source'].nil?
  object.notice_fingerprinter.attributes['source'] == SiteConfig::CONFIG_SOURCE_SITE
end

File: app/views/apps/_fields.html.haml [hunk 3]
```diff
@@ -37,6 +37,39 @@
...
       .field-helpertext times an error occurs (comma separated).
 
+%fieldset
+  %legend= t('.notice_fingerprinter')
+  %p= t('.notice_fingerprinter_text')
+  .checkbox
+    = check_box 'other', 'use_site_fingerprinter', {:checked => app_decorate.use_site_fingerprinter}, SiteConfig::CONFIG_SOURCE_SITE, SiteConfig::CONFIG_SOURCE_APP
+    = f.label :use_site_fingerprinter, t('.use_site_fingerprinter')
+
+  %div.custom_notice_fingerprinter{:style => app_decorate.custom_notice_fingerprinter_style}
```
stevecrozz: This might represent a potential partial view that could be shared with the site-wide configuration area.

config/locales/en.yml
```diff
@@ -183,6 +183,7 @@ en:
       notifications: Notifications
       notify_on_errors: Notify on errors
       notify_all_users: Send notifications to all users
+      notice_fingerprinter_text: The notice fingerprinter governs how error notifications are grouped. Each item counts toward an error's uniqueness if enabled.
```
stevecrozz: This text already exists elsewhere, maybe we should give it a key that can be shared

app/controllers/apps_controller.rb
```diff
@@ -146,6 +149,19 @@ def parse_notice_at_notices_or_set_default
     end
   end
 
+  def process_fingerprinter_choice
+    if params['other'] && params['other']['use_site_fingerprinter']
+      fingerprinter = params['other']['use_site_fingerprinter']
+      if fingerprinter == SiteConfig::CONFIG_SOURCE_SITE
+        app.update_attributes(
+            notice_fingerprinter: SiteConfig.document.notice_fingerprinter_attributes)
```
crunis: You are right, I've switched `write_attributes`. As indicated in http://www.davidverhasselt.com/set-attributes-in-activerecord/ , `write_attributes` doesn't trigger a database access. I've tested and works as claimed. Accessing and modifying app.notice_fingerprinter directly doesn't work because MongoId doesn't get aware of the change and doesn't save it when executing `app.save`.

crunis:
All changes applied. Feel free to review and comment. I'm already working on some tests.

crunis:
Et voilà les tests!

stevecrozz:
This is good. Thanks for all the work on this @crunis !

PR #1135, opened 2016-10-27
Fix backtrace line decorator

badmanski:
Prevent raising error when `object[:file]` is nil

### DIFF
--- a/app/decorators/backtrace_line_decorator.rb
+++ b/app/decorators/backtrace_line_decorator.rb
@@ -2,7 +2,8 @@ class BacktraceLineDecorator < Draper::Decorator
   EMPTY_STRING = ''.freeze
 
   def in_app?
-    object[:file].match Backtrace::IN_APP_PATH
+    return false if file.blank?
+    file.match Backtrace::IN_APP_PATH
   end
 
   def number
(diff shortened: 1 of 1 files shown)


stevecrozz:
@badmanski This seems ok, but how do you have a backtrace without a file? Is that a valid use case? If not, should we implement this as a validation at the time we're receiving the notice?

tartinesKiller:
Hello,
I've encountered this issue with the official Airbrake c# notifier ([https://github.com/airbrake/SharpBrake](url)); the file_name field is set to null by the notifier when the backtrace line is from a lambda expression. (see this issue: [Airbrake](https://github.com/errbit/errbit/issues/1185)). Airbrake seems to accept null filename, so in order to make errbit compliant with Airbrake API it should accept and handle correctly null file_name.

daedeloth:
Merge merge merge! :)

stevecrozz:
This makes sense to me and it looks like some folks are eager to have this. Any chance we can get a couple new tests? Covering these kinds of cases with tests helps a lot when we get to refactoring to make sure we don't end up with the same bug again.

stevecrozz:
Merged the fix under #1260 with tests added.

Issue #1137, opened 2016-11-12
No documentation for running errbit using HTTPS


ccouzens:
There is no documentation for running errbit using https.

By reading the puma documentation, I figured out it could be done by running:

```bash
puma -b "ssl://0.0.0.0:443?key=/certificates/server.key&cert=/certificates/server.crt" -C config/puma.default.rb
```

Would you accept a pull request to add this to the documentation?

Additionally, would you accept a pull request to create an alternative docker image that serves using https?

stevecrozz:
Formerly, we've been doing HTTPS with a separate front end web server to handle the TLS stuff, but if puma can do it, maybe that's just fine for small errbit deployments. A good place for that info would be https://github.com/errbit/errbit/blob/master/docs/deployment.md .

I don't think this would require a separate docker image, although we could explore that if you think it does. Docker users should be able to mount key and cert files onto the Docker container's filesystem and reference them with something like:
```bash
docker run \
  -e "RACK_ENV=production" \
  -e "MONGO_URL=mongodb://my-mongo-host" \
  -v /certificates /host/certificates/path \
  errbit/errbit:latest \
  bundle exec puma -b "ssl://0.0.0.0:443?key=/certificates/server.key&cert=/certificates/server.crt" -C config/puma.default.rb
```

For docker docs, there's probably more value in documenting that people can provide their own arguments to puma (in general) rather than making it all about TLS. But of course, one great example of that is exactly what you're doing by adding a TLS binding and I think there's room in the docs for an example like this in the Docker section: https://github.com/errbit/errbit/blob/master/docs/deployment/docker.md

Does that make sense? I'd like to focus on the flexibility available in general, but I think your example would make for a nice illustration of that.

stevecrozz:
Resolved with https://github.com/errbit/errbit/issues/1139

PR #1139, opened 2016-11-17
Document how errbit can be served on HTTPS

ccouzens:
This is a pull request in response to #1137 .

I haven't created a separate docker image or changed the docker documentation.

-----------------

The section could potentially be broken out into a subpage.

I wanted to link to a blog or article about using a reverse-proxy to
wrap a app-server with HTTPS. Unfortunately, I couldn't find an example
that explained both reverse-proxying and TLS.


stevecrozz:
That's ok. I think setting up a reverse proxy is outside the scope of what what we need to document here. It's more than enough to mention it in passing.

stevecrozz:
Thanks @ccouzens

PR #1144, opened 2016-12-05
fix minor table width problem

weh:
fixes https://github.com/errbit/errbit/issues/1051 without twitter bootstrap

is there a plan to use twitter bootstrap or material design or something like that?


stevecrozz:
There's been a few attempts at twitter bootstrap, but they've all suffered from issues. I'll take a look at this. One of the struggles with fixed width tables in errbit is that the number of columns can change depending on notification services, and issue trackers, etc.

rud:
Probably fixed by  #1217

PR #1145, opened 2016-12-09
Slack channels

gizotti:
This is a follow up PR to #1095.

This adds the ability to users to select a channel in which the Slack notifications will be send when setting up Slack notifications.
The channel field is not compulsory and, if empty, the default channel setup on slack when creating the webhook.

##### TODO
- [x] Write specs to ensure the right payload is sent with a channel both selected or not
- [x] Validate `room_id` field to ensure it matches a slack channels name format

### DIFF
--- a/app/views/apps/_service_notification_fields.html.haml
+++ b/app/views/apps/_service_notification_fields.html.haml
@@ -16,7 +16,8 @@
         %div.notification_params{:class => (w.object.is_a?(notification_service) ? 'chosen ' : '') << notification_service.label}
           - notification_service::FIELDS.each do |field, field_info|
             = w.label field, field_info[:label] || field.to_s.titleize
-            %em= field_info[:hint]
+            - if field_info[:hint].present?
+              %em= field_info[:hint]
             - field_type = field == :password ? :password_field : :text_field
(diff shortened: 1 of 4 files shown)


gizotti:
@stevecrozz #1095 can be closed 😄

app/views/apps/_service_notification_fields.html.haml
```diff
@@ -16,6 +16,7 @@
         %div.notification_params{:class => (w.object.is_a?(notification_service) ? 'chosen ' : '') << notification_service.label}
           - notification_service::FIELDS.each do |field, field_info|
             = w.label field, field_info[:label] || field.to_s.titleize
+            %em= field_info[:hint]
```
stevecrozz: Shouldn't this only be emitted if there is a hint?

gizotti: I guess an if statement here wouldn't hurt. That said, if the `hint` is blank the `em` will have 0 `height`.

stevecrozz:
Thanks @gizotti. This looks good. Will merge soon.

gizotti:
@stevecrozz no worries mate :) Happy to help

PR #1147, opened 2016-12-15
Default to using MongoDB with compression

tamaloa:
MongoDB 3.2 defaults to WiredTiger as storage engine. As with WiredTiger MongDB finally supports compression the resulting database in many cases has a far smaller file size.
If the mongo:3.2 container is started with an existing volume created by the old 3.0 version the old storage engine is automatically used, therefor this change should be downward compatible.


stevecrozz:
Good idea

PR #1148, opened 2016-12-16
Modified API v1 to be able to query problems by date

SValkanov:
Modified API v1 (stats_controller) to return all problems
Modified API v1 (stats_controller) to accept 'after_date' parameter which would return the problems after given date

Modified App model #unresolved_count and #problem_count to accept optional parameter query

### DIFF
--- a/app/models/app.rb
+++ b/app/models/app.rb
@@ -173,6 +173,10 @@ class App
     @problem_count ||= problems.where(query).count
   end
 
+  def problems_notices(query = {})
+    @notices_count ||= problems.where(query).map(&:notices_count).sum
+  end
+
   # Compare by number of unresolved errs, then problem counts.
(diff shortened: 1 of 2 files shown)


app/models/app.rb
```diff
@@ -165,12 +165,16 @@ def copy_attributes_from(app_id)
     end
   end
 
-  def unresolved_count
-    @unresolved_count ||= problems.unresolved.count
+  def unresolved_count(query = {})
+    @unresolved_count ||= problems.unresolved.where(query).count
```
rud: Do note the caching in `@unresolved_count` does not take the `query` parameter into account, which means calling this method with different values for `query` will result in always returning the first computed response. Is this going to be a problem?

rud: Bump

PR #1149, opened 2016-12-19
smtp config: support additional ssl related keys

mattscilipoti:
- smtp_enable_starttls_auto, smtp_openssl_verify_mode

I did not see a way to have a default value for these settings.  I would like to ensure that "enable_starttls_auto" as the expected default value of `true`.

See: http://api.rubyonrails.org/classes/ActionMailer/Base.html#class-ActionMailer::Base-label-Configuration+options

### DIFF
--- a/config/initializers/action_mailer.rb
+++ b/config/initializers/action_mailer.rb
@@ -2,12 +2,14 @@
 if Errbit::Config.email_delivery_method == :smtp
   ActionMailer::Base.delivery_method = :smtp
   ActionMailer::Base.smtp_settings = {
-    address:        Errbit::Config.smtp_address,
-    port:           Errbit::Config.smtp_port,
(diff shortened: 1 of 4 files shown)


stevecrozz:
Good idea @mattscilipoti. Would you also add docs for the new inputs to https://github.com/errbit/errbit/blob/master/docs/configuration.md ?

mattscilipoti:
Docs were being added as you were asking for it.  :)
Any ideas for the default values?

stevecrozz:
I think we oughtta leave the defaults unset because ActiveSupport has its own defaults for these config options. Are you ready for me to merge it?

mattscilipoti:
Yes.  Ready to merge.

stevecrozz:
Thanks @mattscilipoti !

Issue #975, opened 2015-10-16
App Search by name


lucianosousa:
As I mentioned in another issues, it's quite hard to find app infos when you have a lot of them.

I'm thinking about a search by name feature.
What you guys think about it?

stevecrozz:
@lucianosousa how many apps do you have?

lucianosousa:
more than 40 registered.
few are the same but with different envs(acceptance, production)

stevecrozz:
It sounds like that's small enough to do a nice little typeahead client-side search. Something to consider I think as long as people don't tend to have hundreds or thousands of apps.

lucianosousa:
Will you reject my PR if I build it? :)

stevecrozz:
I would pull it in as long as the implementation is simple and has a test or two to make sure it doesn't break.

nielspetersen:
We've included this issue in the current issue (13th) of the RubyIssues: https://rubyissues.ongoodbits.com/. :gem:

scicco:
Hi guys, this is my PR for this, let me know if it's ok.
Thank you in advance for your work!

#1150

PR #1150, opened 2016-12-20
App search by name

scicco:
- search field added in apps list page


scicco:
PR for #975

db/migrate/20161219162030_add_index_into_app.rb
```diff
@@ -0,0 +1,10 @@
+require 'mongoid'
```
stevecrozz: We can probably skip this migration. This should be covered in the upgrade instructions:
https://github.com/errbit/errbit#upgrading

spec/controllers/apps_controller_spec.rb
```diff
@@ -1,5 +1,5 @@
 describe AppsController, type: 'controller' do
-  it_requires_authentication
+  it_requires_authentication for: { search: :get }
```
stevecrozz: I don't understand the need for this change, can you explain?

db/migrate/20161219162030_add_index_into_app.rb
scicco: I've added this migration because i've added a new index named text in App model used here: https://github.com/scicco/errbit/blob/f28a38830be523939e92d1d5b06268a2a6a60760/app/models/app.rb#L57 , so if you have an existing application you must generate the App indexes again to use it.

spec/controllers/apps_controller_spec.rb
scicco: Again you're right @stevecrozz no need of it. I'll remove it.

stevecrozz:
Raised issue #1157 to track the auto-index idea. Merging this one.

Issue #1069, opened 2016-03-01
Unable to clean all issues.


ngw:
I'm dealing with an errbit installation that went south.
The problem is that the database is becoming too big as developers don't resolve the issues, and basically use errbit just to inspect errors they see in production.
I would like now to cleanup the db starting with the errbit:db:clear_resolved I see in my rake tasks. Unfortunately the task starts and never comes back, with some weird error messages I'm not really able to understand.

Overwriting existing field _id in class App.

then a series of

Creating scope :page. Overwriting existing method NotificationServices::CampfireService.page.

I wrote a similar tasks to resolve all elements older than 2 weeks:

    task :cleanup => :environment do
      offset = 3.weeks.ago
      Problem.where(:updated_at.lt => offset).map(&:resolve!)
      Notice.where(:updated_at.lt => offset).destroy_all
    end

Fails for exactly the same reason.

How can I solve this? I could rake db:drop but I don't want this issue to bite me in the future.

scicco:
Hi,

i've fixed this problem by moving all the destroy operations of Problems inside a background job, by introducing Active Jobs with sucker_punch in particular, as already suggested in #1084 for another task.

Here you find my changes: #1151

Hope this helps!

PR #1151, opened 2016-12-20
moving destroy of problems into background jobs

scicco:
- moving destroy of problems into background jobs to avoid #1069 issues

### DIFF
--- a/app/controllers/problems_searcher.rb
+++ b/app/controllers/problems_searcher.rb
@@ -25,6 +25,10 @@ module ProblemsSearcher
       Array(Problem.find(err_ids))
     end
 
+    expose(:selected_problems_ids) do
+      selected_problems.map(&:id).map(&:to_s)
+    end
+
     expose(:err_ids) do
       (params[:problems] || []).compact
     end
--- a/app/jobs/clean_all_problems_job.rb
+++ b/app/jobs/clean_all_problems_job.rb
@@ -1,0 +1,8 @@
+class CleanAllProblemsJob < ActiveJob::Base
+  queue_as :default
+
+  def perform(app_id)
+    app = App.find_by(id: app_id)
+    ::ProblemDestroy.execute(app.problems)
+  end
+end
--- a/app/jobs/clean_several_problems_job.rb
+++ b/app/jobs/clean_several_problems_job.rb
@@ -1,0 +1,12 @@
+class CleanSeveralProblemsJob < ActiveJob::Base
+  queue_as :default
+
+  def perform(problem_ids)
+    bson_problem_ids = []
+    problem_ids.each do |id|
+      bson_problem_ids << BSON::ObjectId.from_string(id)
+    end
+    problems = Problem.find(bson_problem_ids).to_a
+    ::ProblemDestroy.execute(problems)
+  end
+end
--- a/config/application.rb
+++ b/config/application.rb
@@ -48,5 +48,7 @@ module Errbit
 
     # Configure Devise mailer to use our mailer layout.
     config.to_prepare { Devise::Mailer.layout 'mailer' }
+
+    config.active_job.queue_adapter = :sucker_punch
   end
 end
--- a/config/initializers/sucker_punch.rb
+++ b/config/initializers/sucker_punch.rb
@@ -1,2 +1,1 @@
 require 'sucker_punch/async_syntax'
-SuckerPunch.logger = Logger.new("#{Rails.root}/log/sucker_punch.log")
--- a/config/initializers/sucker_punch.rb
+++ b/config/initializers/sucker_punch.rb
@@ -1,0 +1,2 @@
+require 'sucker_punch/async_syntax'
+SuckerPunch.logger = Logger.new("#{Rails.root}/log/sucker_punch.log")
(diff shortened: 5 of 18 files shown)


stevecrozz:
This is excellent! I have a few questions and minor suggestions that I'd like to make, but overall this approach is exactly what I was hoping for (no new dependencies, leaves open the possibility of using a heavy-weight background processor later).

config/initializers/sucker_punch.rb
```diff
@@ -0,0 +1,2 @@
+require 'sucker_punch/async_syntax'
+SuckerPunch.logger = Logger.new("#{Rails.root}/log/sucker_punch.log")
```
stevecrozz: I'm not sure about this one. People configure their loggers in various ways. I think a better default would be to use Rails.logger here. That would make it more likely that the logging configuration people are already using would work.

app/controllers/problems_searcher.rb
```diff
@@ -25,6 +25,10 @@ module ProblemsSearcher
       Array(Problem.find(err_ids))
     end
 
+    expose(:selected_problems_ids) do
+      selected_problems.map(&:id).map(&:to_s)
```
stevecrozz: This is probably fine, but I'm curious about why we need to_s here. Do you know why?

config/application.rb
```diff
@@ -48,5 +48,7 @@ class Application < Rails::Application
 
     # Configure Devise mailer to use our mailer layout.
     config.to_prepare { Devise::Mailer.layout 'mailer' }
+
+    config.active_job.queue_adapter = :sucker_punch
```
stevecrozz: I love this. Active job seems like a really nice abstraction.

spec/jobs/clean_all_job_spec.rb
```diff
@@ -0,0 +1,35 @@
+require 'spec_helper'
```
stevecrozz: This should not be required because its part of the .rspec configuration file: https://github.com/errbit/errbit/blob/v0.6.1/.rspec#L2

app/jobs/clean_all_problems_job.rb
```diff
@@ -0,0 +1,8 @@
+class CleanAllProblemsJob < ActiveJob::Base
```
stevecrozz: Or I guess if the convention is to append Job, it could be `DestroyProblemsByAppJob`

app/jobs/clean_all_problems_job.rb
```diff
@@ -0,0 +1,8 @@
...
+  queue_as :default
+
+  def perform(app_id)
+    app = App.find_by(id: app_id)
+    ::ProblemDestroy.execute(app.problems)
```
scicco: Because without it you get this error on log:
A copy of xxx has been removed from the module tree but is still active
a quick search on google of problem will give this answer:
http://stackoverflow.com/questions/29636334/a-copy-of-xxx-has-been-removed-from-the-module-tree-but-is-still-active (see  Alex Moore-Niemi reply). I tried it and it worked.

app/jobs/clean_all_problems_job.rb
scicco: ok no problem.

app/jobs/clean_several_problems_job.rb
```diff
@@ -0,0 +1,12 @@
+class CleanSeveralProblemsJob < ActiveJob::Base
```
scicco: ok no problem

config/initializers/sucker_punch.rb
scicco: maybe i can comment row 2 and leave the .gitkeep on log folder or you prefer that i'll remove both?
The convention of sucker_punch gem is to log on rails logger as you suggested.

spec/jobs/clean_all_job_spec.rb
scicco: ok i'll remove them

app/controllers/problems_searcher.rb
scicco: This is needed because if you try passing an Array of BSON::ObjectId objects you get 
ActiveJob::SerializationError: Unsupported argument type: BSON::ObjectId

ActiveJob::SerializationError:
       Unsupported argument type: BSON::ObjectId

with this stacktrace
     .../gems/activejob-4.2.5.2/lib/active_job/arguments.rb:69:in `serialize_argument'
     .../gems/activejob-4.2.5.2/lib/active_job/arguments.rb:58:in `block in serialize_argument'
     .../gems/activejob-4.2.5.2/lib/active_job/arguments.rb:58:in `map'

So the simplest way is calling to_s, which gave you a String representation of that ObjectId

and build a BSON::ObjectId with the method from_string

scicco:
Hi @stevecrozz, i've pushed all the changes that you requested. Hope this could be ok. 

Thank you.

app/controllers/problems_searcher.rb
stevecrozz: ah that makes sense. this way job info can be serialized for out-of-process background workers like sideqik, etc.

stevecrozz:
This is wonderful. Thanks @scicco!

PR #1152, opened 2016-12-20
update Dockerfile

chulkilee:
- use ruby:2.3.3-alpine image
- add EXPOSE
- install deps first to leverage layer cache for dev


stevecrozz:
Seems like a good idea to me. Thanks!

PR #1155, opened 2017-01-02
Update README link to Github app registration

giovannibonetti:
When I followed the current link to Github I couldn't find the button to register a new app.
So I think a more specific link would avoid needing to Google it.


stevecrozz:
Very nice.

PR #1162, opened 2017-01-24
Fix tmp directory permissions in Dockerfile

rjnienaber:
Ordering of commands is important in the Dockerfile because when asset:precompile is run, it's done as root, which then creates the /app/tmp file with root permissions. 

Subsequently, after the docker image is started, when connecting to the website for the first time you get a Permission denied error from Devise like this:

image

This relates to the recent changes from #1160


stevecrozz:
Thanks @rjnienaber!

PR #1167, opened 2017-02-14
Add production_only for SlackService

Ceda:
Send only production errors to slack

### DIFF
--- a/app/views/apps/_service_notification_fields.html.haml
+++ b/app/views/apps/_service_notification_fields.html.haml
@@ -15,12 +15,18 @@
       - NotificationService.subclasses.each do |notification_service|
         %div.notification_params{:class => (w.object.is_a?(notification_service) ? 'chosen ' : '') << notification_service.label}
           - notification_service::FIELDS.each do |field, field_info|
-            = w.label field, field_info[:label] || field.to_s.titleize
+            - field_type = field_info[:type] || :text_field
+            - value = field == :notify_at_notices ? w.object.notify_at_notices.join(", ") : w.object.send(field)
+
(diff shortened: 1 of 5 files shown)


spec/models/notification_service/slack_service_spec.rb
```diff
@@ -110,4 +112,15 @@
...
+
+  context 'with only_production' do
+    let(:production_only) { true }
+    let(:notice) { Fabricate :notice, server_environment: { 'environment-name' => 'development' } }
+
+    it "should ignore a notification to Slack" do
+      expect(HTTParty).to_not receive(:post)
+
+      service.create_notification(problem)
+    end
```
rud: Another test of the opposite case would make a lot of sense here - one where ` 'environment-name' => 'production'`, to actually test the flow works as intended.

app/views/apps/_service_notification_fields.html.haml
```diff
@@ -15,12 +15,18 @@
...
+            - if field_type == :check_box
+              %div.checkbox
+                = w.send field_type, field, :placeholder => field_info[:placeholder], :value => value
+                = w.label field, field_info[:label] || field.to_s.titleize
+            - else
+              = w.label field, field_info[:label] || field.to_s.titleize
+              = w.send field_type, field, :placeholder => field_info[:placeholder], :value => value
             - if field_info[:hint].present?
               %em= field_info[:hint]
-            - field_type = field == :password ? :password_field : :text_field
```
rud: This mapping of `:password` to `:password_field` went away, was this intentional?

PR #1174, opened 2017-03-02
Fix notice fingerprinter when backtrace is nil

PikachuEXE:
Sometimes it's nil
Due to storage limit (especially on Heroku)


stevecrozz:
Would you be willing to contribute a test case to make sure this doesn't break again in the future?

PikachuEXE:
Sure, give me some time :)

PikachuEXE:
@stevecrozz 
Spec added and build passed
Please check :)

stevecrozz:
@PikachuEXE Very excellent. Thanks!

PR #1175, opened 2017-03-02
Update gem mongo

PikachuEXE:
This fixes issue like
```
ArgumentError: comparison of Fixnum with nil failed
/app/vendor/bundle/ruby/2.3.0/gems/mongo-2.2.1/lib/mongo/protocol/message.rb:117:in `>'
```
which is fixed in later versions
https://github.com/mongodb/mongo-ruby-driver/commit/0e6fd78c9d6c6c0cf50876172636fa2f4920d54d


stevecrozz:
Looks good. Thanks.

PR #1176, opened 2017-03-08
Expand Clickable Area on Buttons

MicahBrown:
Right now, some of the main navigational buttons are deceptively clickable. I kept finding myself clicking multiple times on the edges of buttons with nothing happening and thinking something wasn't working. It was only because the links have wrapping elements that produce the boundary, which the links don't fully expand across vertically. 

Here's some before/after screens of the css diff: 

| | Main Nav | Action Buttons | Session Nav |
| ----- | ----- | ----- | ----- |
| Before | before | before | before |
| After | after | after | after |


stevecrozz:
Awesome. And thanks for the fantastic before/after screenshots!

PR #1178, opened 2017-03-09
Add spec for email notification

xlu:


stevecrozz:
I'm usually good with adding tests, but does this add coverage beyond what we already get here?
https://github.com/errbit/errbit/blob/3e3a7c3eeaec40eb2b32a9bed87827c7ec010255/spec/models/notice_observer_spec.rb#L46-L57

rud:
@xlu heads up: you leaked a mandrill password in your first commit - hope you have since changed it.

PR #1180, opened 2017-03-14
Deployment

hassan:
This is only "deployment" in the sense of doing a `git pull` from the install directory, but that should be enough for now.


rud:
There seem to be some feature/bug fixes included as well, on top of the stated goal of deployment. If somebody feels up to the task of extracting the various bits we can quite possibly use the work. For now, I'm sorry but I'm going to close this merge request in the interest of keeping a minimal set of work in progress.

PR #1182, opened 2017-03-23
Add support for Ruby 2.4.1

simpl1g:


stevecrozz:
@simpl1g Did this change require updating all these gems? I'd be interested to know which ones needed updating if so. I don't mind updating them all, but if we do that I'd prefer it to be a separate PR.

simpl1g:
@stevecrozz They are all required, I run bundle update gem by gem till tests not passed

stevecrozz:
Interesting! Thanks for the effort.

Issue #1168, opened 2017-02-17
On /users/new the submit button is in spanish


admazzola:
On /users/new the submit button is in spanish using a brand new fresh install.  Everything else is in english.

stevecrozz:
Fixed in #1183

PR #1183, opened 2017-03-28
fix missing english translation

TPei:
Translates submit button on /users/new page to english
See #1168


stevecrozz:
Thanks very much @TPei

PR #1184, opened 2017-03-29
Add backtrace to Slack's notification

### LINKED ISSUE #1104: Add backtrace to Slack's notification
I added a backtrace to my Slack notification.  Is there interest in a PR?

dsandstrom:
It's helpful to see a little more information about the error in Slack,
so now the first five lines of the backtrace show up at the bottom of
the notification.

https://content.screencast.com/users/dsandstrom/folders/Jing/media/ef18cfa1-91f8-4b2d-bd9c-28b0228ae6c4/00000293.png

Fixes #1104

### DIFF
--- a/spec/models/notification_service/slack_service_spec.rb
+++ b/spec/models/notification_service/slack_service_spec.rb
@@ -12,6 +12,15 @@ describe NotificationServices::SlackService, type: 'model' do
   let(:room_id) do
     "#general"
   end
+  let(:backtrace_lines) do
+    backtrace = BacktraceDecorator.new problem.notices.last.backtrace
+    backtrace_lines = ''
+    backtrace.lines[0..4].each do |line|
+      path = line.decorated_path.gsub(%r{</?strong>}, '')
+      backtrace_lines << "#{path}#{line.number} → #{line.method}\n"
+    end
+    "```#{backtrace_lines}```"
+  end
 
   # faraday stubbing
   let(:payload_hash) do
@@ -26,6 +35,7 @@ describe NotificationServices::SlackService, type: 'model' do
           title_link: problem.url,
           text:       problem.where,
           color:      "#D00000",
+          mrkdwn_in:  ["fields"],
           fields:     [
             {
               title: "Application",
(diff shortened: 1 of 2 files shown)


spec/models/notification_service/slack_service_spec.rb
```diff
@@ -12,6 +12,15 @@
   let(:room_id) do
     "#general"
   end
+  let(:backtrace_lines) do
```
rud: Could this please just be an array of strings to test against? That way you avoid duplicating the implementation. 

What do you think?

dsandstrom: I'll give it a try.

dsandstrom: @rud The backtrace is always different.  What strings do you want me to use?  They have to be determined by whatever Fabricate throws at us.

rud: Maybe test for one or more interesting substrings you would expect to see in the output? Something that wasn't there before would be optimal. If the factory is giving you too random data, an idea is usually to more tightly control a few of the things it generates. 

Sorry if that's too abstract, please let me know and I'll help take a look at the concrete behaviour. Deal?

dsandstrom: Do you want me to edit bracktrace_fabricator.rb so it's not random?

dsandstrom: I can do something like

rud: Yeah, wouldn't that make sense in this case?

dsandstrom:
```
app/models/notification_services/slack_service.rb:32:3: C: Metrics/MethodLength: Method has too many lines. [42/36]
  def post_payload(problem) ...
```

Well this method is too big to begin with.  I'm not sure how I can add something to the method without increasing the line count.  The method needs to be broken down into several methods.  Do you want me to refactor it in this PR?

stevecrozz:
Yeah, you don't have to go nuts, but that's the idea. Maybe just pull the `fields` into its own method?

dsandstrom:
Alright, rubocop is happy.  It would be a good idea to let rubocop auto correct and regenerate the todo file.  There a lot of changes since the last time it was generated.

stevecrozz:
@dsandstrom not a bad idea. thanks

PR #1189, opened 2017-04-11
support mentioning users as part of hipchat room notification

sahglie:
When notifications are sent to hipchat rooms, developers can't receive a push notification unless they are @mentioned.  Getting a push notification when an error happens seems like a fairly useful feature: it definitely is for my team.  I've updated the hipchat notification service to include an @mentioned field that will be prepended to the error message so all users @mentioned can receive the push notification.

One caveat about hipchat though, is that their API only allows @mentions when the message format is 'text'.  The default format (and the one currently used by errbit) is 'html'.  So for this pull request, if @mentions is set the notification service will use the 'text' format, otherwise it will keep the original 'html' format.

### DIFF
--- a/app/models/notification_services/hipchat_service.rb
+++ b/app/models/notification_services/hipchat_service.rb
@@ -42,6 +42,8 @@ if defined? HipChat
     end
 
     def create_notification(problem)
+      # @mentions can only be used when format == "text".  See hipchat
+      # api for more info: https://www.hipchat.com/docs/apiv2/method/send_room_notification
       format = self[:mentions].present? ? "text" : "html"
       message = (format == "text") ? message_text(problem) : message_html(problem)
 
--- a/app/models/notification_services/hipchat_service.rb
+++ b/app/models/notification_services/hipchat_service.rb
@@ -17,6 +17,10 @@ if defined? HipChat
       [:room_id, {
         placeholder: "Room name",
         label:       "Room name"
+      }],
+      [:mentions, {
+        placeholder: "@mentions",
+        label:       "@Mentions"
       }]
     ]
     MANDATORY_FIELDS = [:service, :api_token, :room_id]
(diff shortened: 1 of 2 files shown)


app/models/notification_services/hipchat_service.rb
```diff
@@ -38,18 +42,34 @@ def url
     end
 
     def create_notification(problem)
-      url = app_problem_url problem.app, problem
-      message = <<-MSG.strip_heredoc
-        <strong>#{ERB::Util.html_escape problem.app.name}</strong> error in <strong>#{ERB::Util.html_escape problem.environment}</strong> at <strong>#{ERB::Util.html_escape problem.where}</strong> (<a href="#{url}">details</a>)<br>
-        &nbsp;&nbsp;#{ERB::Util.html_escape problem.message.to_s.truncate(100)}<br>
-        &nbsp;&nbsp;Times occurred: #{problem.notices_count}
-      MSG
+      format = self[:mentions].present? ? "text" : "html"
```
stevecrozz: I'm pretty sure everyone will forget why this is necessary. Could you add your note in a comment here?

app/models/notification_services/hipchat_service.rb
```diff
@@ -38,18 +42,34 @@ def url
...
+        <strong>#{ERB::Util.html_escape problem.app.name}</strong> error in <strong>#{ERB::Util.html_escape problem.environment}</strong> at <strong>#{ERB::Util.html_escape problem.where}</strong> (<a href="#{url}">details</a>)<br>
+        &nbsp;&nbsp;#{ERB::Util.html_escape problem.message.to_s.truncate(100)}<br>
+        &nbsp;&nbsp;Times occurred: #{problem.notices_count}
+      MSG
+    end
+
+    def message_text(problem)
+      url = app_problem_url(problem.app, problem)
+      <<-MSG
+#{self[:mentions]} #{problem.app.name} error in #{problem.environment}: #{problem.message.to_s.truncate(100)}
```
stevecrozz: Will this look funny if mentions is nil or blank?

And will anything go wrong for users who have already configured hipchat notifications without mentions and suddenly start using this new code?

app/models/notification_services/hipchat_service.rb
sahglie: done

app/models/notification_services/hipchat_service.rb
sahglie: If mentions is nil or blank (checked via mentions.present?), then the html format will be used.  html format uses the original behavior.  So users that have configured hipchat notifications without mentions will have the same behavior has before.  The only time you get the new behavior is when mentions is set.

stevecrozz: Got it. Thanks @sahglie

stevecrozz:
This is looking good. Normally, I'd insist on adding test coverage for the new code, but we have grand plans to move all of the error notifiers out of errbit and into plugins at some point so I'm thinking of each notifier as its own individual project.

Thanks for contributing!

Issue #955, opened 2015-10-02
SAML authentication


md5:
Would you guys be open to a PR that adds `omniauth-saml` to enable optional SAML authentication for Errbit, similar to how the `GITHUB_AUTHENTICATION` stuff works?

If not, what approach would you recommend for integrating with SAML single-sign on? I'd love it if we didn't have to maintain a fork of Errbit in order to get SAML authentication.

stevecrozz:
SAML integration sounds great. We should just talk about how it can work and be useful to a wide audience of SAML users. Do you have a place I can go look at your implementation?

md5:
@stevecrozz I haven't written any code yet, but I'm going to talk to my colleagues next week about setting aside some time to do so. Since we'd rather not maintain a fork, my thought was that I'd fall back to using GitHub authentication against an internal GitHub Enterprise instance if you guys weren't amenable to adding `omniauth-saml` to Errbit itself.

stevecrozz:
I'm fine with including that gem as long as the implementation is generic enough that other organizations could use it, and especially if you're willing to help maintain this feature and answer questions. I would just want to talk over how exactly the integration would work. Here's a barrage of starting questions do get the ball rolling:

Would you be able to provision new Errbit users using SAML or just log existing users in? New Errbit users would likely need at least an email address. Would we have to assume the name-id assertion is an email address? What if the SAML response doesn't contain any email address assertions? What about other assertions like first/last name or full name? Maybe we can ignore those for now and let users set their own names in the edit profile UI. Will SAML users still be able to log in with a password? Will they even be able to set a password? Would Errbit support SP-initiated login or just IdP-initiated log in? Do we care about logouts?

By the way, which IdP are you using?

md5:
> Would you be able to provision new Errbit users using SAML or just log existing users in? New Errbit users would likely need at least an email address.

Since all apps under Errbit can be seen by all users (correct me if I'm wrong here), I think it would only make sense to allow SAML logins for existing users provisioned in Errbit. The SAML profile would be required to have an email address configured in order to make it available to Errbit for notifications.

I think this would at least work as an initial pass. I could see eventually supporting the mapping of a "role" or "member of" assertion to allow that to be used the way that `GITHUB_ORG_ID` is currently used.

> Would we have to assume the name-id assertion is an email address? What if the SAML response doesn't contain any email address assertions? What about other assertions like first/last name or full name? Maybe we can ignore those for now and let users set their own names in the edit profile UI.

[...]

PR #1192, opened 2017-05-02
SAML Authentication

MarkusHarmsen:
As discussed in #955, a SAML authentication could be useful for some users (at least it is for us).

Here is a proposal, on how to implement SAML authentication. It has been tested with the GSuite.


stevecrozz:
@MarkusHarmsen this looks great. I've been putting this PR off too long and I think its time to focus on it because there's a lot of value here for SAML users.

I don't particularly want the Vagrantfile, but still interested in why you have it. Are you just more familiar with Vagrant? or did you have a problem with the Docker setup?

MarkusHarmsen:
@stevecrozz I removed the Vagrant stuff and updated the pull request.

The reason for Vagrant was just that I'm more familiar with Vagrant than docker ;)

stevecrozz:
This is nice @MarkusHarmsen. One last question for you before I merge this. Are you willing to act as a sponsor for this feature, helping to answer questions and take occasional Errbit tickets relating to SAML auth? I ask because if we merge this, then we're also declaring that we support it. If our users start using this, they'll definitely raise issues and ask questions, and I don't always have the time to answer.

MarkusHarmsen:
Sure - before merging this, let me one time completely re-setup an errbit-instance an configure it using SAML (GSuite). I will document all required steps, so that might be a good starting point for a better documentation.

stevecrozz:
Sounds great @MarkusHarmsen!

stevecrozz:
@MarkusHarmsen it looks like there's another feature that will probably arrive in master soon that will conflict with this one: #1221. I don't think it will be very difficult to resolve, but it seems likely to happen.

stevecrozz:
@MarkusHarmsen This PR is so good and so close, I'd hate to let it fade away. Are you still intending to add more documentation?

PR #1194, opened 2017-05-11
Deploy

cognoauser:
Set up standard Capistrano deployment for staging and production errbit servers.


stevecrozz:
Most likely the wrong target for this PR since it has some site-specific customizations. But I actually enjoy it when this happens because I get to see what folks are doing with Errbit!

hassan:
Doggone it - messed up while working with 2 different GH IDs. Please close with extreme prejudice (and accept my apologies).

stevecrozz:
No problem at all.

PR #1196, opened 2017-05-22
Allow sending email notification for *every* error

floehopper:
This makes the behaviour of `ERRBIT_EMAIL_AT_NOTICES` consistent with that of `ERRBIT_NOTIFY_AT_NOTICES`.

We work hard to keep the exception notifications in our apps to zero. The default configuration is to only send emails for the 1st, 10th and 100th occurrence of an error. We've found that this sometimes means we mark the 1st occurrence error as resolved and then it recurs a few more times and we are not notified. We'd prefer to be notified every time, hence this change.

### DIFF
--- a/app/models/error_report.rb
+++ b/app/models/error_report.rb
@@ -87,8 +87,8 @@ class ErrorReport
 
   # Send email notification if needed
   def email_notification
-    return false unless app.emailable?
-    return false unless should_email?
+    return unless app.emailable?
+    return unless should_email?
     Mailer.err_notification(self).deliver_now
   rescue => e
     HoptoadNotifier.notify(e)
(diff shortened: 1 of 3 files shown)


app/models/error_report.rb
```diff
@@ -83,7 +83,7 @@ def cache_attributes_on_problem
   # Send email notification if needed
   def email_notification
     return false unless app.emailable?
-    return false unless app.email_at_notices.include?(@problem.notices_count)
+    return false unless app.email_at_notices.include?(0) || app.email_at_notices.include?(@problem.notices_count)
```
stevecrozz: I prefer to keep these guard clauses without compound conditionals. Could you put this on two lines?

floehopper: Sure. No problem. I'll do it now and force-push the changed commit.

floehopper:
Sorry - I got distracted. New commits just pushed up. I hope you don't mind, but I've done a bit more tidying up in the new commits to make the two methods as similar as possible. I think that makes things clearer.

stevecrozz:
Very nice. Merging now.

floehopper:
Thanks for merging. Any chance of a release including this change?

PR #1200, opened 2017-05-26
Fix table layout on the problems page (#1051)

mknkisk:
## Summary

See #1051

By not specifying the cell width of `WHERE & WHAT`, auto layout the table.
because if the cell size is fixed, table overflows the width of container when the number of columns changes. (e.g. issue column)

## Before

errbit_current

## After

errbit_fix_layout

### DIFF
--- a/app/assets/stylesheets/errbit.css.erb
+++ b/app/assets/stylesheets/errbit.css.erb
@@ -683,6 +683,9 @@ td.count, td.issue_link {
 }
 
 /* Err Tables */
+table.errs th, table.errs td {
+  word-break: break-all;
+}
 table.errs td.app .name {
   display: block;
   width: 150px;
@@ -696,11 +699,6 @@ table.errs td.app .environment {
 }
 table.errs td.message a {
   display: block;
-  word-wrap: break-word;
-  width: 440px;
-  white-space: nowrap;
-  text-overflow: ellipsis;
-  overflow: hidden;
 }
 table.errs td.message em {
   color: #727272;


stevecrozz:
Thanks for looking into this! This area has been changed a lot over the years and there's always been some problem with it. And you're right about the problem with fixed width columns. For me, the ideal behavior would be to truncate the excess text with text-overflow: ellipsis and using title attributes to show the full text on hover (tooltip).

If there were thousands of characters in that text field, we probably don't want to display them all outside of a tooltip. I believe that was the previous problem before we landed on what we have today. But truncating them naively using CSS does seem to require fixed width columns which do have the problem you're fixing.

But I spent a little time looking around today, and there are some interesting ideas in the non-accepted section on this stackoverflow page. Some of them seem to indicate how we could get table-like behavior with text-overflow control without having to specify widths:
https://stackoverflow.com/questions/9789723/css-text-overflow-in-a-table-cell

In particular, I'm looking at this one which shows one fixed width column and the rest are not fixed. But it seems to work even when the first column is not fixed:
http://jsfiddle.net/maruxa1j/

Any thoughts?

mknkisk:
@stevecrozz Thank you for your response 😄 

I also think that the idea (jsfiddle) is good. However, I am trying it, but it is not going well...

I do not have any idea what to do if there are multiple elements in a cell 🙈 
The `What & Where` cell contains three elements.

1. what message
2. where message
3. last comment (user name & comment body)

---

Other idea

I could apply ellipsis by specifying a width of 300px for all three elements.

long-where-message-layout-fix

I will also consider other ideas :dizzy:

stevecrozz:
Yeah this is tricky. I can have a look soon too, but I do like the way that looks with the static 300px width. I would prefer something more fluid if we can get there. Does it not work when there are multiple cell contents to put each one in its own width: 100% container?

stevecrozz:
Hello @mknkisk!

I have done a bit of research on the problem and created this branch. Would you like to have a look at it? https://github.com/errbit/errbit/compare/master...stevecrozz:1051-table-layout-fix-revisited?w=1

mknkisk:
Hi, @stevecrozz 

Sorry for late reply 🙈 I saw your branch.

And I checked the table layout assuming the case that the column was further increased. The table fits within the container 👍 

I think this method is better than my pull request. (The table layout collapsed on my branch 😓 )

---

💡 💡 💡 

It seems that "issue_link" does not have ".narrow-as-possible" class 👀 

```diff
if any_issue_links
-    %td.issue_link
+    %td.issue_link.narrow-as-possible
```

Before:



After:

stevecrozz:
Thanks for taking a look @mknkisk. I'll close this one and we can continue the conversation in #1217.

PR #1201, opened 2017-06-11
Added message patterns (to group errors together with same error mess…

keikeicheung:
…age after replacing regex pattern)

### DIFF
--- a/app/assets/stylesheets/errbit.css.erb
+++ b/app/assets/stylesheets/errbit.css.erb
@@ -329,6 +329,11 @@ form textarea {
   font-size: inherit; font-family: inherit;
   color: #787878; border: 1px solid #C6C6C6;
 }
+
+form div.message_patterns textarea {
+  width: 96%;
+  height: 80px;
+}
 form textarea.short { height: 8em; }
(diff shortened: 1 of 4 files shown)


stevecrozz:
This is an interesting idea. Can you describe the real-world case you're solving? I'd like to learn about something concrete that you could do with this.

keikeicheung:
Ahh, sorry forgot to put in more description. i.e. like if the error msg contains some sort of id - like "failed to process order id: xxxxxxxx for some job" then instead of grouping when whole error message string matches, you can put in order id: (\S*) pattern to identify the xxxxxxx is the order id - which can be replaced by "*" then all orders that failed to be processed for the job will be grouped together. Because in some case, errors sent to errbit are what they are logging in the logs. So in the logs I still want the order id to be shown in the error msg, but that won't affect the grouping (does this make sense?)

stevecrozz:
This does seem useful in the abstract.

But I'm motivated to keep Errbit as small as possible, and because the same thing can be done in the client (https://github.com/airbrake/airbrake-ruby#airbrakeadd_filter), I think its best to do this on the client side. Did you try that approach and reject it already?

PR #660, opened 2014-02-19
Outdated Error cleaner

deathowl:
The requested modifications are applied to the code, and also separated in a different workbranch

PR #1204, opened 2017-06-29
Feature outdated problem clearer

hennevogel:
Rebase of #660

Now that the deployment tracking is gone I could really use a way to clear old `Problems` without clicking...

### DIFF
--- a/lib/tasks/errbit/database.rake
+++ b/lib/tasks/errbit/database.rake
@@ -11,7 +11,7 @@ namespace :errbit do
   end
 
   desc "Delete old errors from the database. (Useful for limited heroku databases)"
-  task :clear_outdated => :environment do
+  task clear_outdated: :environment do
     require 'outdated_problem_clearer'
     if Errbit.config.errbit_problem_destroy_after_days.present?
       puts "=== Cleared #{OutdatedProblemClearer.new.execute} outdated errors from the database."
--- a/lib/tasks/errbit/database.rake
+++ b/lib/tasks/errbit/database.rake
@@ -10,6 +10,16 @@ namespace :errbit do
     puts "=== Cleared #{ResolvedProblemClearer.new.execute} resolved errors from the database."
   end
 
+  desc "Delete old errors from the database. (Useful for limited heroku databases)"
+  task :clear_outdated => :environment do
+    require 'outdated_problem_clearer'
+    if Errbit.config.errbit_problem_destroy_after_days.present?
+      puts "=== Cleared #{OutdatedProblemClearer.new.execute} outdated errors from the database."
+    else
+      puts "=== ERRBIT_PROBLEM_DESTROY_AFTER_DAYS not set. Old problems will not be destroyed."
+    end
+  end
+
   desc "Regenerate fingerprints"
   task notice_refingerprint: :environment do
     NoticeRefingerprinter.run
--- a/app/interactors/outdated_problem_clearer.rb
+++ b/app/interactors/outdated_problem_clearer.rb
@@ -1,0 +1,31 @@
+require 'problem_destroy'
+
+class OutdatedProblemClearer
+  ##
+  # Clear all problem not present for more than one week.
+  #
+  def execute
+    nb_problem_outdated.tap do |nb|
+      if nb > 0
+        criteria.each do |problem|
+          ProblemDestroy.new(problem).execute
+        end
+        repair_database
+      end
+    end
+  end
+
+private
+
+  def nb_problem_outdated
+    @count ||= criteria.count
+  end
+
+  def criteria
+    @criteria ||= Problem.where(:last_notice_at.lt => Errbit::Config.notice_deprecation_days.to_f.days.ago)
+  end
+
+  def repair_database
+    Mongoid.default_client.command repairDatabase: 1
+  end
+end
--- a/app/interactors/outdated_problem_clearer.rb
+++ b/app/interactors/outdated_problem_clearer.rb
@@ -22,7 +22,7 @@ private
   end
 
   def criteria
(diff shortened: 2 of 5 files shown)


stevecrozz:
This seems like a good idea @hennevogel. I've got a few change requests that I'll make in the PR but overall this seems like a good solution for folks who like to keep a tidy house.

.env.default
```diff
@@ -25,3 +25,4 @@ GITHUB_SITE_TITLE=GitHub
 DEVISE_MODULES='[database_authenticatable,recoverable,rememberable,trackable,validatable,omniauthable]'
 GOOGLE_AUTHENTICATION=true
 GOOGLE_SITE_TITLE=Google
+ERRBIT_NOTICE_DEPRECATION_DAYS=7
```
stevecrozz: Let's leave this turned off for now by not including it in .env.default. I'd also like to change the word from deprecated to something a bit more scary sounding because this does actually cause `problems` to be destroyed (in a good way). Maybe `ERRBIT_PROBLEM_DESTROY_AFTER_DAYS`

app/interactors/outdated_prolem_clearer.rb
```diff
@@ -0,0 +1,32 @@
...
+  def nb_problem_outdated
+    @count ||= criteria.count
+  end
+
+  def criteria
+    @criteria ||= Problem.outdated
+  end
+
+  def repair_database
+    Mongoid.default_session.command :repairDatabase => 1
```
stevecrozz: This is probably fine. I can't remember why or if this is needed, but I see we have it elsewhere so I guess we probably should do exactly what you have done unless we figure out that answer.

app/models/problem.rb
```diff
@@ -61,6 +61,7 @@ class Problem
   scope :ordered, -> { order_by(:last_notice_at.desc) }
   scope :for_apps, ->(apps) { where(:app_id.in => apps.all.map(&:id)) }
   scope :search, ->(value) { where('$text' => { '$search' => value }) }
+  scope :outdated, where(:last_notice_at.lt => Errbit::Config.notice_deprecation_days.to_f.days.ago)
```
stevecrozz: I think its not worth having this named scope to use it in just one place.

lib/tasks/errbit/database.rake
```diff
@@ -10,6 +10,12 @@ namespace :errbit do
     puts "=== Cleared #{ResolvedProblemClearer.new.execute} resolved errors from the database."
   end
 
+  desc "Delete old errors from the database. (Useful for limited heroku databases)"
+  task :clear_outdated => :environment do
+    require 'outdated_problem_clearer'
+    puts "=== Cleared #{OutdatedProblemClearer.new.execute} outdated errors from the database."
```
stevecrozz: This seems good. It might be nice to include some more informational output in a switch here to help guide people to this feature if they're looking for it. For example:
```ruby
if Errbit.config.errbit_problem_destroy_after_days.present?
  puts "=== Cleared #{OutdatedProblemClearer.new.execute} outdated errors from the database."
else
  puts "=== ERRBIT_PROBLEM_DESTROY_AFTER_DAYS not set. Old problems will not be destroyed."
end

File: .env.default (hunk 1 above)
Comment (stevecrozz) [inline .env.default]: And once the name and default is settled, this should be documented in https://github.com/errbit/errbit/blob/master/docs/configuration.md

File: app/interactors/outdated_prolem_clearer.rb [hunk 5]
```diff
@@ -0,0 +1,32 @@
...
+  def execute
+    nb_problem_outdated.tap { |nb|
+      if nb > 0
+        criteria.each do |problem|
+          ProblemDestroy.new(problem).execute
+        end
+        repair_database
+      end
+    }
+  end
```
stevecrozz: I'd like to add one test here just to execute at least one success case. You can use a mock to set the config temporarily since it won't have a default setting. Something like
```ruby
allow(Errbit::Config).to receive(:errbit_problem_destroy_after_days).and_return(30)
```

stevecrozz (changes requested):
Thanks for this. I've left a number of requests in here and I know I'm probably asking a lot for free but I don't mind helping out if you need.

Thanks for contributing!

app/interactors/outdated_prolem_clearer.rb
hennevogel: Dunno, @deathowl might...

app/interactors/outdated_prolem_clearer.rb
hennevogel: I've added a spec similar to the `ResolvedProblemClearer` spec. Did not find the time to test this locally, let's see 🤠

stevecrozz:
Looks like the specs are crashing, may need to require the problem cleaner file in the spec. I'm going to approve the changes though because they otherwise look good. Just need specs to pass.

stevecrozz (approved):

hennevogel:
@stevecrozz what do you mean by 'include'? The interactors are in the environment, which is loaded....

stevecrozz:
@hennevogel I'm talking about the CI failure where we see `uninitialized constant OutdatedProblemClearer `. The report is available on travis at https://travis-ci.org/errbit/errbit/builds/250428596

hennevogel:
@stevecrozz that's what I mean with `The interactors are in the environment, which is loaded`. But there was a typo in the filename and rails only loads classes called like the filename. Stupid me... fixed.

stevecrozz:
@hennevogel Almost there! Travis is reporting a few errors in the build, but the build is running now:
https://travis-ci.org/errbit/errbit/jobs/261822806

Can you check it out?

hennevogel:
@stevecrozz updated...

stevecrozz:
Looks good @hennevogel! Thanks for the contribution and for spending the effort required to make this a part of Errbit. I'm sure others will appreciate it too.

PR #1207, opened 2017-07-12
Add more Google features

smerrill:
- Allow automatic account creation as with GitHub
- Allow limiting which Google domains can create accounts
- Add documentation of the flags involved

### DIFF
--- a/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/users/omniauth_callbacks_controller.rb
@@ -40,17 +40,28 @@ class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
   end
 
   def google_oauth2
+    google_auto_register = Errbit::Config.google_auto_register
+    google_domain = Errbit::Config.google_domain
     google_uid = env['omniauth.auth'].uid
     google_email = env['omniauth.auth'].info.email
     google_user = User.where(google_uid: google_uid).first
     google_site_title = Errbit::Config.google_site_title
+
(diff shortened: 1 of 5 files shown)


smerrill:
This shouldn't affect existing installs - so long as neither `GOOGLE_AUTO_REGISTER` or `GOOGLE_DOMAIN` are set, Google accounts should work as they always have.

stevecrozz:
Very nice @smerrill! I know this code area is a little light on unit tests. Would you be able to add a bit of coverage? Breaking up google_oauth2 might help with that.

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -40,17 +40,28 @@ def github
...
+    google_auto_register = Errbit::Config.google_auto_register
+    google_domain = Errbit::Config.google_domain
     google_uid = env['omniauth.auth'].uid
     google_email = env['omniauth.auth'].info.email
     google_user = User.where(google_uid: google_uid).first
     google_site_title = Errbit::Config.google_site_title
+
+    if google_domain && !google_email.end_with?("@#{google_domain}")
+      # Fail if the email does not match our domain
+      flash[:error] = "'#{google_email}' is unauthorized. Only #{google_domain} email addresses are allowed."
```
pquerner: Why disclose this information?

rud: Good point, should be minimal disclosure. @smerrill you up for fixing this?

PR #1213, opened 2017-07-28
Apply system timezone to slack notification

tazu-yyy:
I changed to use system time zone in Slack notification. Before this change, Slack notification was sending the "First Noticed" time as UTC in any configuration.

I think it is better to use user settings. Could you please check these changes?


rud:
Looks like a neat idea. Would you mind rebasing / fixing the merge conflict?

PR #1214, opened 2017-08-11
restricted access mode

sebastianhegge:
Implementing restricted access mode to realize user to app relation managed only by admin.
Admin can watch + unwatch users in app.
If mode is activated, users can only see and access apps and errors they are watching.


stevecrozz:
Thanks for the PR @sebastianhegge. I know that access control is a feature that some users would like to have, but I believe it should be a separate concept from 'watchers'. I'm not sure how much effort you want to put into designing and building this feature, but if you would like, we could start a new issue for brainstorming ideas on how access control could work or we could chat about it in gitter...

sebastianhegge:
Thanks @stevecrozz for your comment.
I know that user rights and the watching feature are some different functions.
The feature to limit some useraccess to some apps is for us for data protection reasons a compelling feature. And we don't have the need to give an user access to an app - but the user don't want to watch the app.
So this is the fastest and most practical solution for us.

stevecrozz:
It makes sense to me, and I appreciate the offer, but I don't wish to pull it into Errbit master with watcher and access concepts combined.

sebastianhegge:
From conceptual point of view i can understand your decision. Even though I think this is better than nothing - as long as there is no other solution.
Thanks for reviewing.

PR #1215, opened 2017-08-14
#fix-ui-collapse | when the env info is too long,the UI will collapse :(

iChenLei:
>errbit.version = 0.6 && 0.7-dev
### Hi,the team of errbit,when I use errbit to collect errors,I found a issue 💥 about UI as follows.
collapse
### This is a problem just because of the **css** file and I fixed it.
errbit_2_

### DIFF
--- a/app/assets/stylesheets/errbit.css.erb
+++ b/app/assets/stylesheets/errbit.css.erb
@@ -692,12 +692,14 @@ table.errs td.app .name {
 }
 table.errs td.app .environment {
   font-size: 0.8em;
+  word-break:break-all;
+  text-overflow:ellipsis;
   color: #999;
 }
 table.errs td.message a {
   display: block;
   word-wrap: break-word;
-  width: 440px;
+  width: 400px;
   white-space: nowrap;
   text-overflow: ellipsis;
   overflow: hidden;


stevecrozz:
Thanks for the report @iChenLei. There is another problem that you may not have seen, and that is the number of columns can change depending on Errbit's configuration. There is another pull request that's a bit further along and I've also made some recommendations. Would you like to read the conversation in this other PR and help discuss? https://github.com/errbit/errbit/pull/1200

iChenLei:
@stevecrozz  Oops, I ignore the special situation. Yeah, I found the difference between my pr and PR-1200, my table has 6 columns and mknkisk's table has 7 columns. My change can't be suited to PR-1200.  we haven't the best way to solve it yet? That's bad..  But for me, this is not a big problem, I think the 【repeated error report】is more important.Now I am working for it , look the screenshot as follows
wee
【repeated error report】is so annoying,and make us love errbit and hate errbit...
So we want to do something make it be acceptable.Hope you can find a better unique error judge strategy.I am glad to communicate with you..(From shanghai china, 00：33 🌔 )

stevecrozz:
Sure. I'd love to have a conversation about error fingerprinting. I think that's what you're wanting to discuss. I'm assuming you have found the area under 'config' where you can fine tune the way fingerprinting works. Do you have some examples of errors that you'd like to have grouped that are not grouped the way you would like?

iChenLei:
@stevecrozz   Thank you,You help me solve a big problem,The right config can group same errors, It's my fault..I am an intern in SAP.Before I start my internship,The errbit in our production has been running a long time.My leader ignore the config and use the default configuration. And now I need create a new dashboard page and let it show all user's issue(bug), just like JIRA issues page.I feel it's not easy to do.If I meet any problems,can I asking for help? Thank you all.

stevecrozz:
Ah. If this is an old install, it may have been set up before it was possible to configure the fingerprinting. Anyway, I'm glad its working for you, and good luck!

PR #1216, opened 2017-08-30
Fix webhook service notification payload

pioneerskies:
NotificationServices::WebhookService#message_for_webhook
previously returned an Hash with a single key "problem"
with a string as value:

{
    "problem": "Stringified json here"
}

actually loosing almost all the "problem" properties.

Now the "problem" key bring on a child Hash as expected.

### DIFF
--- a/app/models/notification_services/webhook_service.rb
+++ b/app/models/notification_services/webhook_service.rb
@@ -15,7 +15,7 @@ class NotificationServices::WebhookService < NotificationService
 
   def message_for_webhook(problem)
     {
-      problem: { url: problem.url }.merge(problem.as_json).to_json
+      problem: { url: problem.url }.merge(problem.as_json)
     }
   end
 
--- a/spec/models/notification_service/webhook_service_spec.rb
+++ b/spec/models/notification_service/webhook_service_spec.rb
@@ -9,4 +9,16 @@ describe NotificationServices::WebhookService, type: 'model' do
 
     notification_service.create_notification(problem)
   end
+
+  context "#message_for_webhook" do
+    it "should return an hash with all the keys nested inside 'problem' key" do
+      notice = Fabricate :notice
+      notification_service = Fabricate :webhook_notification_service, app: notice.app
+      problem = notice.problem
+
+      payload = notification_service.message_for_webhook(problem)
+
+      expect(payload[:problem]).to be_a(Hash)
+    end
+  end
 end


pioneerskies:
As an extra explanation this is the previously generated paylod:

```json
{
    "problem": "{\"url\":\"http:\/\/errbit.welaika.com\/apps\/56586820b95851000c00000b\/problems\/59a6c7add6561b00070001a9\",\"_id\":\"59a6c7add6561b00070001a9\",\"app_id\":\"56586820b95851000c00000b\",\"app_name\":\"Prasco\",\"comments_count\":0,\"created_at\":\"2017-08-30T16:11:57.775+02:00\",\"environment\":\"production\",\"error_class\":\"ActiveRecord::RecordNotFound\",\"first_notice_at\":\"2017-08-30T16:11:57.772+02:00\",\"hosts\":{\"a2a5453aae95b8c0f81d4fc244406429\":{\"value\":\"www.prasco.net\",\"count\":1}},\"issue_link\":null,\"issue_type\":null,\"last_notice_at\":\"2017-08-30T16:11:57.786+02:00\",\"message\":\"Couldn't find PartCompatibility with 'id'=100024\",\"messages\":{\"08ea2640793cce6d8189881cf6c3aebf\":{\"value\":\"Couldn't find PartCompatibility with 'id'=100024\",\"count\":1}},\"notices_count\":1,\"resolved\":false,\"resolved_at\":null,\"updated_at\":\"2017-08-30T16:11:57.775+02:00\",\"user_agents\":{\"6231b6afc78725894ba9924cc1293b4f\":{\"value\":\"Mozilla 5.0 (bingbot\/2.0)\",\"count\":1}},\"where\":\"part_compatibilities#show\"}"
}
```

now we get a plain Hash. My only doubt is how `as_json` problem's method is handling DateTime objects, but I've tested the returnign json and it LGTM.

Thanks

stevecrozz:
This does seem like an area we need to address. But I do have a concern that folks may have built integrations that will suddenly break after this change. We could either decide to not care about that and forge ahead, or perhaps make a second webhook notifier.

Either way, it also seems like a good idea to revisit the message format and make sure it looks the way we want because if we're going to make a backwards incompatible change, we may as well take the opportunity to break as much of it as we want so that we only have to do it once.

What do you think?

pioneerskies:
Yours are good points.

I think the backward or not compatibility matter is something only a maintainer could afford to decide about. As demonstrated, I have not the wider picture in mind :)

What I think is that only custom code should work with actual webhook implementation: in order to use that data you have to receive it, get the value of the main node (`problem` key) and `JSON.parse` it.

From my point of view "WebHook notices" should be an agnostic, standard way to interact with majority of destinations; including services such as Zapier's webhook listeners or Ryver's ingoing webhooks, just to say something I'm on with. And there you can't transform data so much. And overall i think about a general named WebHook as a POST sending out JSON; full stop. Receiving a stringyfied data structure should be unattended, imho.

The idea to leave the old `NotificationServices::WebhookService` alone and write a new class, sounds good. But I dunno how to evidently underline and explain their differences in code and in UI.

I'm at least happy to have emerged a thoughtful discussion.

Thank you very much for your attention and feedback.

stevecrozz:
I've thought this over a bit more and you're of course right about the way this should work, but you're also right to recommend the breaking change. I think it would be more confusing to have two webhook notifiers, one with correct behavior and one with backwards-compatible behavior. So that means we may break a few existing integrations. That does bother me a bit, but I'm hoping people will be able to make the changes to adapt without too much trouble.

PR #1217, opened 2017-08-31
allow fluid table columns on problem index

stevecrozz:
Ever since we started allowing tables to have varying numbers of columns, we've had problems with our static layout. There are a number of competing goals we have:

1. tables should look fine even when there's lots of non-breaking characters in each of the text fields
2. when text is too long, it should `overflow: ellipsis`
3. when text is too long, it should have a title tag with (more) complete text on hover
4. when text is too long, we should avoid excessively tall table cells, refusing to wrap text wherever possible
5. when columns are added/removed via configuration, the table should remain as wide as its container and not blow outside its bounds
6. ideally, the table should still behave like a classic HTML table and adapt dynamically to the size of its contents

This is surprisingly difficult, but I think we've got something here.

### DIFF
--- a/app/assets/stylesheets/errbit.css.erb
+++ b/app/assets/stylesheets/errbit.css.erb
@@ -652,7 +652,7 @@ td.message .line {
 td.latest {
   white-space: nowrap;
 }
-td.count, td.issue_link {
+td.count .td-container, td.issue_link .td-container, td.resolve .td-container {
   text-align: center;
 }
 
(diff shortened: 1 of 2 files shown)


stevecrozz:
@mknkisk points out that issue_link is missing .narrow-as-possible. Planning to add that.

stevecrozz:
It's looking pretty good to me. I am a little annoyed that the column widths seem to be calculated based on the text inside `<thead>` but overall I think this is an improvement.

Issue #1190, opened 2017-04-20
Feature Request: Support snoozing notifications


MohamedBassem:
I have an errbit app with slack notifications enabled. Whenever a cronjob fails, it gets restarted and I get a slack notification from errbit which is nice. When the job is restarted and fails again, I get another slack message. The channel is then spammed with the same error happening over and over again even when I'm actually fixing the bug. My feature request is adding the ability to snooze/silence the notifications of a certain error for a period of time (while the issue is being fixed for example). The only workaround I found was to disable the notifications of the whole app, which is not optimal.

rud:
Closing this as #1218 was closed. Have a look at `ERRBIT_NOTIFY_AT_NOTICES` to see if that would help you stem the flow.

PR #1218, opened 2017-09-03
Add snooze feature for problem's new notices

Farideh:
according to #1190

### DIFF
--- a/app/assets/stylesheets/errbit.css.erb
+++ b/app/assets/stylesheets/errbit.css.erb
@@ -742,6 +742,18 @@ table.tally th.value {
   margin-right: 5px;
 }
 
+/* Snooze Errs */
+#action-bar .snooze:before {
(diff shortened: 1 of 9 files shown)


stevecrozz:
I have a few questions about this. Firstly, I want to make sure that you have already seen ERRBIT_NOTIFY_AT_NOTICES and that it doesn't solve the problem at hand. Secondly, this feature here uses the name 'snooze' which to me implies some kind of timer like the snooze feature of an alarm clock. Maybe you snooze now and the snooze becomes active, but then later the snooze automatically deactivates. I think what is implemented here is more like a 'mute' which is also interesting in its own way.

But before we make any changes, I would like to hear about your experience with ERRBIT_NOTIFY_AT_NOTICES because I think its designed to help with the stated problem in #1190. Perhaps all we need is better documentation or a better default value for this configuration.

PR #1219, opened 2017-09-03
Fix portuguese word in en.yml

2called-chaos:
Used to be called "Tenant" in older versions, now it's portuguese, I propose "Origin"?


stevecrozz:
Thanks @2called-chaos!

PR #1221, opened 2017-09-07
Improve Google Auth

Ginja:
* Add ARG variables to allow for UID/GID mapping
* Update MAINTAINER (deprecated) instruction to LABEL
* Improve HEALTHCHECK, now curls the default IP of the container
* Add GOOGLE_AUTO_PROVISION config option to toggle automatic account creation
* Add GOOGLE_REDIRECT_URI config option to set a custom redirect location. This is useful if errbit is being served behind a reverse proxy that's terminating TLS
* Add GOOGLE_AUTHORIZED_DOMAINS config option to whitelist trusted domains for login
* Allow for custom user during database seed

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,11 @@
 FROM ruby:2.3.3-alpine
-MAINTAINER David Papp <david@ghostmonitor.com>
+LABEL maintainer="David Papp <david@ghostmonitor.com>"
 
-RUN addgroup -S errbit \
-  && adduser -S -D -s /bin/false -G errbit -g errbit errbit
+ARG UID=101
+ARG GID=101
+
+RUN addgroup -g $GID -S errbit \
+  && adduser -u $UID -S -D -s /bin/false -G errbit -g errbit errbit
 
 # throw errors if Gemfile has been modified since Gemfile.lock
 RUN echo "gem: --no-document" >> /etc/gemrc \
@@ -45,6 +48,6 @@ RUN chown -R errbit:errbit /app
 
 USER errbit
 
-HEALTHCHECK CMD curl --fail http://localhost:$PORT/users/sign_in || exit 1
+HEALTHCHECK CMD curl --fail "http://$(/sbin/ip route | /usr/bin/awk '/src/{print $NF}'):8080/users/sign_in" ||  exit 1
 
 CMD ["bundle","exec","puma","-C","config/puma.default.rb"]
--- a/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/users/omniauth_callbacks_controller.rb
@@ -70,7 +70,7 @@ class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
           redirect_to new_user_session_path, alert: user.errors.full_messages.join("\n")
         end
       else
-        flash[:error] = "Account's email domain is not authorized for login. Please log in with an account from a trusted domain."
+        flash[:error] = I18n.t "devise.google_login.domain_unauthorized"
         redirect_to new_user_session_path
       end
     else
--- a/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/users/omniauth_callbacks_controller.rb
@@ -62,8 +62,8 @@ class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
     elsif Errbit::Config.google_auto_provision
       user = User.create_from_google_oauth2(request.env['omniauth.auth'])
       if user.persisted?
-        flash[:notice] = I18n.t "devise.omniauth_callbacks.success", :kind => google_site_title
-        sign_in_and_redirect user, :event => :authentication
+        flash[:notice] = I18n.t "devise.omniauth_callbacks.success", kind: google_site_title
+        sign_in_and_redirect user, event: :authentication
       else
         session['devise.google_data'] = request.env['omniauth.auth'].except(:extra)
         redirect_to new_user_session_path, alert: user.errors.full_messages.join("\n")
--- a/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/users/omniauth_callbacks_controller.rb
@@ -50,7 +50,7 @@ class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
       if google_user && google_user != current_user
         flash[:error] = "User already registered with #{google_site_title} login '#{google_email}'!"
       else
-        # Add github details to current user
+        # Add google details to current user
         current_user.update(google_uid: google_uid)
         flash[:success] = "Successfully linked #{google_email} account!"
       end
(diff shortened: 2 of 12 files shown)


Ginja:
Will cleanup the code, & add tests. Just putting this out there now for feedback or concerns.

stevecrozz:
Hey @Ginja. Can you write a bit about how you're testing the Docker image? It seems clear that containers are taking over the world and I'd like to get to a place where we're a bit more specific about the interface we're providing with our Docker image. At some point, I'd really love to write some regression tests around this Dockerfile. That's maybe not necessary right now, but I'd be interested to know why you're making these Dockerfile changes and how you're making sure they are doing what you want.

Second thing is I'd like to know about your use case for GOOGLE_AUTO_PROVISION. Do you really want anyone in the world with a google account to be able to use your Errbit?

One last question is about GOOGLE_REDIRECT_URI. Do you have a case in mind where you'd want that to be something other than the dashboard of the Errbit deployment? Meaning is it worth having the option to use something other than the value of ERRBIT_HOST?

Ginja:
Hey @stevecrozz, sorry for the delay.

I'm using an internal tool for testing the Docker image, but it's nothing that couldn't be accomplish with Rake. These are the Docker commands I'm using:

```
docker build -t errbit:latest .

docker run --rm -p 127.0.0.1:27017:27017 --name errbit-mongo -d mongo:3.2

docker run --rm -e ERRBIT_ADMIN_USER=admin -e ERRBIT_ADMIN_EMAIL=errbit@example.com -e ERRBIT_ADMIN_PASSWORD=password -e RACK_ENV=production -e MONGO_URL="mongodb://docker.for.mac.localhost:27017" errbit rake errbit:bootstrap

# Environment variables are just specific to what I cared about and of course can easily be changed
docker run --rm -p 127.0.0.1:8080:8080 --name errbit -e RACK_ENV=production -e ERRBIT_HOST=errbit.example.com -e ERRBIT_PROTOCOL=http -e ERRBIT_PORT=80 -e GOOGLE_AUTHENTICATION=true -e GOOGLE_AUTO_PROVISION=true -e GOOGLE_CLIENT_ID=$CLIENT_ID -e GOOGLE_SECRET=$CLIENT_SECRET -e GOOGLE_REDIRECT_URL="http://127.0.0.1:8080/users/auth/google_oauth2/callback" -e GOOGLE_AUTHORIZED_DOMAINS="errbit.example.com" -e MONGO_URL="mongodb://docker.for.mac.localhost:27017" errbit:latest

# Shut down commands

docker kill errbit-mongo

docker kill errbit
```

The changes I made are strictly for ease-of-use in production. 

### Dockerfile

If you want to terminate TLS within Errbit then as the [docs](https://github.com/errbit/errbit/blob/master/docs/deployment.md#https) state you need to serve it with some extra parameters specifying the cert & key. This usually means you need to bundle the certs within the container or mount them at runtime. However, there's a good chance the `errbit` user inside the container won't be able to read the files as the UID/GID won't map to an appropriate user outside of the container. The `ARG` statements allow users to build a container with a specific UID/GID that matches a user on their system, making it easier to pass in files that the `errbit` user can read.

From my tests, the existing healthcheck would always fail as the $PORT variable wasn't being interpolated. And from what I could tell, the HEALTHCHECK directive doesn't accept variables. So I changed it to the most reasonable default. It's probably best to change this in the future to run a script instead that can make smarter decisions.

The MAINTAINER to LABEL change was just [future-proofing](https://docs.docker.com/engine/reference/builder/#maintainer-deprecated) :)

### GOOGLE_AUTO_PROVISION & GOOGLE_AUTHORIZED_DOMAINS

We run Errbit in a restricted environment so if you can hit it, we want you to have access. And in configuring auto-provisioning, we reduce the operational toil of needing to create you an account beforehand. That being said, you bring up a _very_ good point for other use-cases. Which is I also added GOOGLE_AUTHORIZED_DOMAINS, which can be a comma-delimited list of email domains you're willing to accept logins from.

### GOOGLE_REDIRECT_URI

The redirect URI defaults to HTTP, and if you're not terminating TLS on the Errbit container (e.g. reverse-proxy scenario), it can cause issues. This config let's you specify a custom location so you don't need to configure your web server to do HTTP->HTTPS redirection (you should anyways). Even though the spec doesn't require you to redirect to a HTTPS location, I feel better in doing so.

Dockerfile
```diff
@@ -45,6 +48,6 @@ RUN chown -R errbit:errbit /app
 
 USER errbit
 
-HEALTHCHECK CMD curl --fail http://localhost:$PORT/users/sign_in || exit 1
+HEALTHCHECK CMD curl --fail "http://$(/sbin/ip route | /usr/bin/awk '/src/{print $NF}'):8080/users/sign_in" ||  exit 1
```
stevecrozz: I think this is good. I can't think of any reason to need to change the bound port when running in a container.

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -59,6 +59,20 @@ def google_oauth2
...
+        user = User.create_from_google_oauth2(request.env['omniauth.auth'])
+        if user.persisted?
+          flash[:notice] = I18n.t "devise.omniauth_callbacks.success", kind: google_site_title
+          sign_in_and_redirect user, event: :authentication
+        else
+          session['devise.google_data'] = request.env['omniauth.auth'].except(:extra)
+          redirect_to new_user_session_path, alert: user.errors.full_messages.join("\n")
+        end
+      else
+        flash[:error] = "Account's email domain is not authorized for login. Please log in with an account from a trusted domain."
```
stevecrozz: Ideally, these messages would use I18n.t so that they could be translated more easily in the future. I can merge without doing that, but it is nice to have.

app/models/user.rb
```diff
@@ -47,6 +47,27 @@ class User
     validates :username, presence: true
   end
 
+  def self.valid_google_domain?(email)
+    return true if Errbit::Config.google_authorized_domains.nil?
+    match_data = /.+@(?<domain>.+)/.match(email)
```
stevecrozz: This is probably fine since the `@` anchors the position, but I think it would be better to ensure this is the end of the string `/.+@(?<domain>.+)$/`

app/models/user.rb
```diff
@@ -47,6 +47,27 @@ class User
...
+    return false if match_data.nil?
+    Errbit::Config.google_authorized_domains.split(",").include?(match_data[:domain]) ? true : false
```
stevecrozz: There's no need for the ternary because the return value of `include?` here is already correct.

db/seeds.rb
```diff
@@ -4,9 +4,9 @@
 puts "-------------------------------"
 
 # Create an initial Admin User
-admin_username = "errbit"
-admin_email = "errbit@#{Errbit::Config.host}"
-admin_pass  = SecureRandom.urlsafe_base64(12)[0, 12]
+admin_username = ENV['ERRBIT_ADMIN_USER'] || "errbit"
```
stevecrozz: This seems like a great idea. Not sure why I didn't think of it.

stevecrozz:
Thanks @Ginja. I'm just about ready to merge this. I made a couple suggestions that I would consider optional. But take a look and see what you think and let me know if you're ready. Sadly, I think #1192 will conflict, but this PR may arrive in master first.

Just so you know, when users report issues with this feature I'll probably assign them to you :) Are you ok with that?

Ginja:
Yup, feel free.

PR in with fixes! Good catch on that redundant ternary.

PR #1224, opened 2017-09-08
Fix Ruby 2.4 Capistrano Deployment

felixbuenemann:
This fixes exceptions in net-ssh and sshkit that occured when trying to deploy on MRI Ruby 2.4.1 and updates capistrano to 3.4.1 for rake 11.0 compatibility.

**Note:** Because of the capistrano update, users will need to bump the `lock` version in `config/deploy.rb` from `3.4.0` to `3.4.1`.


stevecrozz:
Looks good to me. I seem to recall that there was some other problem with running Ruby 2.4, but I can't remember the specifics. Maybe this was it...

PR #1226, opened 2017-09-12
Improve the notices query in comment notification

aquajach:
Decorator's association like `ProblemDecorator.new(comment.err).notices.first` doesn't optimise the query, instead it pulls all the `notices` from database. When a `problem` has many `notices`, this becomes a critical performance issue. 

Perhaps no additional test is required.


stevecrozz:
Thanks @aquajach. This seems reasonable and something I'll try to watch out for.

PR #1227, opened 2017-09-19
Fix #1225 - Use X-Airbrake-Token header when present

### LINKED ISSUE #1225: Support the X-Airbrake-Token header
I've been trying the [new Java airbrake notifier](https://github.com/airbrake/javabrake) however they are sending the API key via a `X-Airbrake-Token` header instead of as a query parameter (see [here](https://github.com/airbrake/javabrake/blob/9953317c70fe2307d91b8573a7588b6ea5db8611/src/main/java/OkSender.java#L53)). Errbit doesn't seem to support that. Is this a difficult one to implement? :)

alexmreis:
Fixes compatibility issues with javabrake, the new Airbrake java client. 

This client uses the X-Airbrake-Token header instead of sending the apiKey in the ?key=foobar query parameter.


stevecrozz:
Thanks. This looks ok to me.

PR #1237, opened 2017-11-08
Allow Authorization header for authentication

### LINKED ISSUE #1236: Support key passed through Authorization header
As of airbrake-ruby 2.5.1 the project key is passed through the Authorization header instead of as a query param.  (https://github.com/airbrake/airbrake-ruby/pull/278)

excid3:
This fixes #1236 where airbrake-ruby 2.5.1 added the Authorization header for the project key instead. 

See airbrake/airbrake-ruby#278

### DIFF
--- a/app/controllers/api/v3/notices_controller.rb
+++ b/app/controllers/api/v3/notices_controller.rb
@@ -29,7 +29,7 @@ class Api::V3::NoticesController < ApplicationController
     render text: 'Invalid request', status: 400
   end
 
-  private
+private
 
   def authorization_token
     request.headers['Authorization'].to_s[/Bearer (.+)/, 1]
--- a/app/controllers/api/v3/notices_controller.rb
+++ b/app/controllers/api/v3/notices_controller.rb
@@ -14,7 +14,7 @@ class Api::V3::NoticesController < ApplicationController
 
     merged_params = params.merge(JSON.parse(request.raw_post) || {})
     merged_params = merged_params.merge('key' => request.headers['X-Airbrake-Token']) if request.headers['X-Airbrake-Token']
-    merged_params = merged_params.merge('key' => authorization_token) if request.headers['Authorization']
+    merged_params = merged_params.merge('key' => authorization_token) if authorization_token
     report = AirbrakeApi::V3::NoticeParser.new(merged_params).report
 
     return render text: UNKNOWN_API_KEY, status: 422 unless report.valid?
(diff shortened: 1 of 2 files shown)


stevecrozz:
Nice! Just a couple of things. There's a couple of minor rubocop failures, could use a test, and then I'm also looking at the regex wondering what happens if an Authorization header is provided that doesn't match. It looks like it would be a nil reference error that we could avoid.

app/controllers/api/v3/notices_controller.rb
```diff
@@ -27,4 +28,11 @@ def create
   rescue AirbrakeApi::ParamsError
     render text: 'Invalid request', status: 400
   end
+
+  private
+
+  def authorization_token
```
tindron: How about something like this to avoid the nil reference?

```
def authorization_token
  /Bearer (?<key>.+)/ =~ request.headers['Authorization']

  key
end
```

app/controllers/api/v3/notices_controller.rb
```diff
@@ -14,6 +14,7 @@ def create
 
     merged_params = params.merge(JSON.parse(request.raw_post) || {})
     merged_params = merged_params.merge('key' => request.headers['X-Airbrake-Token']) if request.headers['X-Airbrake-Token']
+    merged_params = merged_params.merge('key' => authorization_token) if request.headers['Authorization']
```
tindron: How about making the condition `if authorization_token` so it doesn't replace the key if it isn't present in the Authorization header?

app/controllers/api/v3/notices_controller.rb
excid3: I was thinking similar. This makes it a nice one-liner.

```ruby
def authorization_token
  request.headers['Authorization'][/Bearer (.+)/, 1]
end
```

kuboon: @excid3 It looks fail if `request.headers['Authorization']` is nil

excid3:
Added tests for both the Authorization head and a malformed value. 👍

stevecrozz:
Looks perfect. You rock!

app/controllers/api/v3/notices_controller.rb
```diff
@@ -14,6 +14,7 @@ def create
 
     merged_params = params.merge(JSON.parse(request.raw_post) || {})
     merged_params = merged_params.merge('key' => request.headers['X-Airbrake-Token']) if request.headers['X-Airbrake-Token']
+    merged_params = merged_params.merge('key' => authorization_token) if authorization_token
```
chabgood: It would be faster if you set the param directly:
merged_params['key'] = authorization_token) if authorization_token

mileslane:
@ndbroadbent, any chance of getting a new release?  We have been hoping to see one since November.  Thanks.

PR #1238, opened 2017-11-14
Speed up common idioms

chabgood:
Replaced ruby and rails idioms for faster code.  https://github.com/JuanitoFatas/fast-ruby

### DIFF
--- a/app/controllers/api/v3/notices_controller.rb
+++ b/app/controllers/api/v3/notices_controller.rb
@@ -13,8 +13,9 @@ class Api::V3::NoticesController < ApplicationController
     return render(status: 200, text: '') if request.method == 'OPTIONS'
 
     merged_params = params.merge(JSON.parse(request.raw_post) || {})
-    merged_params = merged_params.merge('key' => request.headers['X-Airbrake-Token']) if request.headers['X-Airbrake-Token']
-    merged_params = merged_params.merge('key' => authorization_token) if authorization_token
+    # merge makes a copy, merge! edits in place
+    merged_params = merged_params.merge!('key' => request.headers['X-Airbrake-Token']) if request.headers['X-Airbrake-Token']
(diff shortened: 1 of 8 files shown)


app/controllers/api/v3/notices_controller.rb
```diff
@@ -13,8 +13,9 @@ def create
     return render(status: 200, text: '') if request.method == 'OPTIONS'
 
     merged_params = params.merge(JSON.parse(request.raw_post) || {})
-    merged_params = merged_params.merge('key' => request.headers['X-Airbrake-Token']) if request.headers['X-Airbrake-Token']
-    merged_params = merged_params.merge('key' => authorization_token) if authorization_token
+    # merge makes a copy, merge! edits in place
+    merged_params = merged_params.merge!('key' => request.headers['X-Airbrake-Token']) if request.headers['X-Airbrake-Token']
+    merged_params = merged_params.merge!('key' => authorization_token) if authorization_token
```
stevecrozz: Hey, this all looks great @chabgood. I think in this one spot there's probably no need to assign the return value of `merge!` to `merged_params` as that's done automatically by `merge!`. Is that right?

app/controllers/problems_controller.rb
```diff
@@ -102,7 +102,7 @@ def merge_several
   end
 
   def unmerge_several
-    all = selected_problems.map(&:unmerge!).flatten
+    all = selected_problems.flat_map(&:unmerge!)
```
stevecrozz: Well there's something I've never used. Thanks for pointing it out.

app/controllers/api/v3/notices_controller.rb
chabgood: merged_params is just a variable above.  I edited the code above.

PR #1242, opened 2017-12-12
adding health check endpoints. w/ burnettk

jasquat:
Adding health check endpoints. We need this in order to configure load balancers and orchestration tools like kubernetes.


stevecrozz:
Great PR! Will merge when travis passes.

PR #1249, opened 2018-01-04
CI: upgrade ruby-versions

rud:


rud:
So, turns out an upgrade of Devise is at least necessary before ruby 2.5.0 can even boot the app. Keeping it in the test-matrix will at least highlight this fact

stevecrozz:
@rud do you think its time to drop support for Ruby 2.2.x? Maintenance for this version ends in March. And I suppose we no longer need 2.5 in the allowed failures area, right?

rud:
Well, devise needs a major upgrade before the app will even boot on ruby 2.5.0, but removing ancient 2.2 certainly makes sense.

stevecrozz:
@rud looking at this build: https://travis-ci.org/errbit/errbit/jobs/325554155

It looks like Ruby 2.5 passes (aside from a rubocop problem). What issue with Devise did you find? Is it related to this PR? https://github.com/errbit/errbit/pull/1251

PR #1250, opened 2018-01-05
Travis CI: use specific mongodb database versions

rud:
Setup from https://docs.travis-ci.com/user/database-setup/#MongoDB

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ cache: bundler
 services:
   - mongodb
 before_script:
+  - until nc -z localhost 27017; do echo "waiting for mongodb"; sleep 1; done
   - bundle exec rake errbit:bootstrap
 script:
   - bundle exec rspec
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,12 +15,10 @@ sudo: false
 
 cache: bundler
 
-install:
+before_script:
   - wget https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-${MONGODB}.tgz
   - tar xzf mongodb-linux-x86_64-${MONGODB}.tgz
   - ${PWD}/mongodb-linux-x86_64-${MONGODB}/bin/mongod --version
-
-before_script:
   - mkdir ${PWD}/mongodb-linux-x86_64-${MONGODB}/data
   - >
     ${PWD}/mongodb-linux-x86_64-${MONGODB}/bin/mongod
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,7 @@ before_script:
     --logpath ${PWD}/mongodb-linux-x86_64-${MONGODB}/mongodb.log
     --auth --fork
   - until nc -z localhost 27017; do echo "waiting for mongodb ${MONGODB}"; sleep 1; done
+  - bundle install
   - bundle exec rake errbit:bootstrap
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,11 +10,9 @@ env:
   - COVERAGE=true JRUBY_OPTS=--debug
 sudo: false
 cache: bundler
+services:
+  - mongodb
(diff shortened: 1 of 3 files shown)


stevecrozz:
We used to use travis' provided mongodb, but there seems to be no way to control which version they provide and they change it whenever they want. I think its important to control the mongo version in order to have a consistent test environment. But certainly willing to hear any arguments if you think there's more benefit to using the provided mongodb.

rud:
@stevecrozz that's an excellent point, and controlling the mongo version certainly isolates from somewhat random CI environment changes. On the other hand it also isolates from database changes as versions are upgraded, but the CI stays with the old database version. 

When looking around I saw [this](https://github.com/mongodb/pymodm/blob/master/.travis.yml) as an elegant way of testing with multiple mongodb versions, might be relevant here?

rud:
@stevecrozz this lowering of ruby-version specificity, what do you think of that change? Should I pull that out into a separate PR for inclusion?

stevecrozz:
@rud what does the ruby-version change do? Do those map to specific ruby versions in travis? or does it mean something like "latest patch version for each minor version"?

Regarding mongodb versions, I think running multiple versions in a matrix is a cool idea, but part of me thinks it may not be necessary. One potential way forward is to declare support for only one version of mongo. I'm shorter on time these days than I used to be and I tend towards strategies that reduce my own maintenance burder :)

That said, errbit does enjoy community support as you're proving right now. And if you want to help support multiple mongo versions, I won't stand in the way.

rud:
The ruby-version change means rvm will use a fuzzy version matcher when activating ruby, so yes, "2.2" means latest 2.2.x version. I think that makes sense to use - and the test output logs what specific version has been activated, so any sudden weird failures should be easy to debug.

For the mongodb versions, one fairly simple thing to do would be to keep the existing test-version, as well as also testing on the latest 3.6 release, matrix environment style, like I linked. That way there'll at least be clarity about what versions are supported. Then again, maybe just picking a single version would be just fine, I'm certainly no mongodb fountain of knowledge.

Also, now that I'm looking at it, maybe it's time to remove the broken java-versions from the Travis config, they often take half an hour to fail.

stevecrozz:
I'm fine with all that, and also fine with removing support for 2.2 at this time as we'll have to remove it soon anyway.

rud:
Aaalright, once CI finishes with this current batch I think I'll stop fiddling with this particular area. MongoDB 3.6.1 is not compatible with the current mongoid version used, so that probably has to be updated at some point.

I think everything else is as we discussed above, @stevecrozz, but please let me know if it's not.

rud:
Regarding the mongodb 3.6.1 failure, it fails like this: https://travis-ci.org/errbit/errbit/builds/325841571

```
  1) ErrorReport with notification service configured send email
     Failure/Error: app.save
     
     Mongo::Error::OperationFailure:
       Unknown modifier: $pushAll (9)
     # ./spec/models/error_report_spec.rb:227:in `block (3 levels) in <top (required)>'
```

rud:
Hmm, am I alone in seeing 3 messages from coveralls for each pushed set of commits?

stevecrozz:
Yeah that's annoying. This happens because we do multiple builds on travis (one per ruby version). I'm sure we could rework that logic so that we only get coverage on a single build.

stevecrozz:
@rud should be an easy fix. I'll open a new issue to track the mongo 3.6 incompatibility. Notes on how to deal with it are here: https://docs.mongodb.com/v3.2/reference/operator/update/pushAll/#up._S_pushAll

stevecrozz:
And actually, and this can be part of the new task, but I think we should stop supporting these old mongodb versions because of the info I'm finding on the mongo support policy doc: https://www.mongodb.com/support-policy

It looks like even 3.0 is going to be end-of-lifed next month.

PR #1251, opened 2018-01-05
Devise: conservative upgrade from 3.5.5 to 3.5.10

rud:
Changelog shows only a few bugfixes:
https://github.com/plataformatec/devise/blob/a0af72edfdf2328bcc6586f57748c30422f2cabc/CHANGELOG.md


stevecrozz:
Sounds like this is change is required in order to support Ruby 2.5, so I'll merge this first.

PR #1258, opened 2018-01-26
Fix rake errbit:clear_outdated task

delphaber:
The configuration was probably renamed in the past, but the rake task and service object was not updated


stevecrozz:
Ouch, and the specs didn't pick it up because we had it mocked. Thanks.

PR #1259, opened 2018-02-01
Fix docker healthcheck for mutiple networks

philipgraf:
If you attach multiple networks to the docker container the `/sbin/ip route | /usr/bin/awk '/src/{print $NF}'` will print something like this:
```
172.18.0.4
172.20.0.3
172.21.0.3
```

And the HEALTHCHECK command will fail with the following:
```
curl: (3) Illegal characters found in URL
```

This fix picks the first IP and uses this for checking if the service is available or not

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -48,6 +48,6 @@ RUN chown -R errbit:errbit /app
 
 USER errbit
 
-HEALTHCHECK CMD curl --fail "http://$(/sbin/ip route | /usr/bin/awk '/src/{print $NF}'):8080/users/sign_in" ||  exit 1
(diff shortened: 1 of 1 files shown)


stevecrozz:
You know, when I run this on one machine, some weird output:
```bash
$ /sbin/ip route | /usr/bin/awk '/src/{print $NF}'
linkdown
172.18.0.1
172.19.0.1
172.172.0.1
600
192.168.99.1
```

Should we consider using `hostname -I` instead?

stevecrozz:
I think this is good, but I'll confess I don't know all the ways people use this. Hopefully this works for everyone.

PR #1262, opened 2018-02-16
Handle empty environment variables

### LINKED ISSUE #1261: Empty environment variable crashes Configurator
If there's an empty environment variable, `YAML.parse` returns `false`, which does not have `to_ruby` method.

Related https://github.com/errbit/errbit/blob/8024da5761a2cf069255957633e03dae129f8118/lib/configurator.rb#L60

For verification:
```
YAML.parse('').to_ruby
```
Generates the error as well.

psukys:
`YAML.parse` for an empty value `''` returns `false` instead of an expected object, thus directly feeding that into `to_ruby` method fails. This handles the case where environment variable is empty.

Fixes #1261

### DIFF
--- a/lib/configurator.rb
+++ b/lib/configurator.rb
@@ -58,11 +58,7 @@ class Configurator
       @overrides[key] = values.pop if values.last.is_a? Proc
       env_name = values.find { |v| ENV[v] }
       if env_name
-        if ENV[env_name].empty?
-          @storage[key] = ''
-        else
-          @storage[key] = YAML.parse(ENV[env_name]).to_ruby
-        end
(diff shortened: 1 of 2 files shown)


psukys:
As I'm not a well versed ruby developer, thus proposed solution might not look as nice, but for me there's more of a question of how this should be handled. I currently just set an empty string.

psukys:
Alright, rubocop helped me to sort out the gruesome if tirade

rud:
Nice work :+1:

rud:
@stevecrozz over to you for final verify

stevecrozz:
Looks good. I had no idea YAML.parse('') == false. Thanks for the contribution.

PR #1266, opened 2018-04-16
 avoid truncating count where it is a four digit number

burnettk:
does not really affect the look to my untrained eye.

before:

image

after:

image

there is no similar easy fix for five-digit numbers. :)

### DIFF
--- a/app/assets/stylesheets/errbit.css.erb
+++ b/app/assets/stylesheets/errbit.css.erb
@@ -659,7 +659,7 @@ td.count .td-container, td.issue_link .td-container, td.resolve .td-container {
 
 .count a {
   display: inline-block;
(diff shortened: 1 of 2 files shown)


burnettk:
update snarky PR title since it's no longer a one-character css change.

stevecrozz:
I'll try to find some time to play with this. This table css is really finicky and small changes have caused all kinds of annoying problems in the past. So I'll need to check the overflow of each column in the various possible table states (there are states where more or fewer columns are present in the table).

stevecrozz:
This seems to work great, actually. I hope I didn't miss anything, but I can't find a good reason to avoid merging this, so in it goes.

PR #1268, opened 2018-04-16
do not show app in table when already on an app/show page

burnettk:
this updates pages like /apps/5a78bd63b33eb400a20e88dd, like this.

before:

image

after:

image

Sorting by environments (which is implied as a feature by having an "ENV" column) now seems to work.

this table partial in question is used by two actions:
 * apps/show
 * problems/index

the conditional (which was already being used elsewhere in the same partial) makes it so problems/index continues to be rendered as before (show the app as well as the env).

this feels slightly higher risk than the other two PRs I just filed, but still very low risk, and hopefully it's a pure UX improvement and not at all controversial. but let me know!

### DIFF
--- a/app/controllers/problems_searcher.rb
+++ b/app/controllers/problems_searcher.rb
@@ -3,7 +3,7 @@ module ProblemsSearcher
 
   included do
     expose(:params_sort) do
-      if %w(app message last_notice_at count).member?(params[:sort])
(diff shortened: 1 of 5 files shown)


stevecrozz:
I'm fine with this, but it looks like the build failed. I think in this case its ok to extend the rubocop limit. At some point we need to break up this class, but I think it doesn't have to be right now.

PR #1269, opened 2018-04-17
fix some nilpointer-style exceptions in the view

burnettk:
when there are no notices attached to problem. new test fails pre-controller-fix.


stevecrozz:
Makes sense. Thanks @burnettk

PR #1270, opened 2018-04-17
kubernetes health checks should return 500 on failure. w/ burnettk

jasquat:
The kubernetes probe system wants non-200's on failures.


stevecrozz:
Agreed. I'm surprised we didn't do this already actually.

PR #1271, opened 2018-04-23
add new health api that can be used to test client config, including api_key

burnettk:
we have existing checks for errbit clients that actually use the normal "send-a-notice" api, but it's a bit suboptimal that errbit then ends up sending emails and whatnot. this new API provides pretty much the same assurance that the errbit client is configured correctly, but with fewer side effects. working with @jasquat .

### DIFF
--- a/app/controllers/health_controller.rb
+++ b/app/controllers/health_controller.rb
@@ -12,6 +12,13 @@ class HealthController < ActionController::Base
     render json: { ok: true }, status: :ok
   end
 
+  def api_key_tester
+    app = App.where(api_key: params[:api_key]).first
+    is_good_result = app ? true : false
+    response_status = is_good_result ? :ok : :error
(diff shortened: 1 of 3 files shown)


app/controllers/health_controller.rb
```diff
@@ -12,6 +12,13 @@ def liveness
     render json: { ok: true }, status: :ok
   end
 
+  def api_key_tester
+    app = App.where(api_key: params[:api_key]).first
+    is_good_result = app ? true : false
+    response_status = is_good_result ? :ok : :error
```
rud: Suggestion: should `:error` not be `:forbidden` here?

rud:
Other than that one HTTP UI detail looks, it looks to me like this code just follows the existing API token checking convention in the codebase.

burnettk:
@rud addressed!

PR #1272, opened 2018-04-29
GitHub Auth: Show error msg on empty email

crunis:
When using GITHUB_ORG_ID to automatically sign up new users, if the user has no email selected in his public profile, i.e, settings/profile looks like this:

image

Errbit won't receive any email and won't be able to create a new user, redirecting back to the sign in page without displaying any message.

This PR makes Errbit show a "Flash Error" message with the text: "You need to define and select a valid email in your Github profile"

### DIFF
--- a/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/users/omniauth_callbacks_controller.rb
@@ -1,20 +1,27 @@
 class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
+  def github_auto_sign_up(github_token)
+    return if Errbit::Config.github_org_id.nil?
+
+    # See if the user is a member of the organization that we have access for
+    # If they are, automatically create an account
+    client = Octokit::Client.new(access_token: github_token)
+    client.api_endpoint = Errbit::Config.github_api_url
+    org_ids = client.organizations.map(&:id)
+    return nil unless org_ids.include?(Errbit::Config.github_org_id)
+
+    if env["omniauth.auth"].extra.raw_info.email.nil? || env["omniauth.auth"].extra.raw_info.email == ""
+      flash[:error] = "You need to define and select a valid email in your Github profile"
+      nil
+    else
+      User.create(name: env["omniauth.auth"].extra.raw_info.name, email: env["omniauth.auth"].extra.raw_info.email)
+    end
+  end
+
   def github
     github_login = env["omniauth.auth"].extra.raw_info.login
     github_token = env["omniauth.auth"].credentials.token
-    github_user  = User.where(github_login: github_login).first
     github_site_title = Errbit::Config.github_site_title
-
-    if github_user.nil? && (github_org_id = Errbit::Config.github_org_id)
-      # See if they are a member of the organization that we have access for
-      # If they are, automatically create an account
-      client = Octokit::Client.new(access_token: github_token)
-      client.api_endpoint = Errbit::Config.github_api_url
-      org_ids = client.organizations.map(&:id)
-      if org_ids.include?(github_org_id)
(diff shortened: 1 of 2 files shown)


stevecrozz:
Hey there @crunis !

Is it possible that we would get access to user's email address if we request the user:email scope by setting `GITHUB_ACCESS_SCOPE='["repo", "user:email"]'`? It seems like we should be able to read the user's private email addresses assuming we have permission:
https://developer.github.com/apps/building-oauth-apps/scopes-for-oauth-apps/#available-scopes

crunis:
Hi @stevecrozz !

Good point, I'll check that out

crunis:
`env["omniauth.auth"]` only contains the user's public profile, so adding only the "user:email" permission doesn't retrieve private emails.

My last commit adds the logic to do so.

stevecrozz:
@crunis I like this change. I see the github interface being mocked. I'm assuming you tested it with a real github integration, yeah?

crunis:
Of course! I tested the whole thing with my Errbit setup. Works as expected!

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -79,10 +89,32 @@ def google_oauth2
...
+      return email.email unless email.nil?
+
+      email = client.emails.first
+      return email.email unless email.nil?
+    rescue Octokit::ClientError => e
+      Rails.logger.warn "Octokit:ClientError exception while retrieving user's emails. We probably lack user:email permission. Will try to extract email from user's public profile. Error message: #{e}"
+    end
+
+    # Try to get email from public profile
+    if !env["omniauth.auth"].extra.raw_info.email.nil? && env["omniauth.auth"].extra.raw_info.email != ""
```
rud: `env["omniauth.auth"].extra.raw_info.email.present?` should do the same

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -1,20 +1,28 @@
...
+    client.api_endpoint = Errbit::Config.github_api_url
+    org_ids = client.organizations.map(&:id)
+    return nil unless org_ids.include?(Errbit::Config.github_org_id)
+
+    user_email = github_get_user_email(client)
+    if user_email.nil?
+      flash[:error] = "Could not retrieve user's email from GitHub"
+      nil
+    else
+      User.create(name: env["omniauth.auth"].extra.raw_info.name, email: user_email)
```
rud: Doesn't this duplicate part of what was just done in the `github_get_user_email` method?

rud:
@crunis I just had a few low-level code details, would you mind taking a look at them?

app/controllers/users/omniauth_callbacks_controller.rb
crunis: Agree! fixed

app/controllers/users/omniauth_callbacks_controller.rb
crunis: I don't see the duplication, github_get_user_email gets the email.
Then, here we create the user using the email and adding the username, where's the duplication?

crunis:
@rud I've addressed one of your comments and commented in the other, do you mind following up?

Sorry for the delay to answer!

rud:
Over to you @stevecrozz for final validation :)

crunis:
Btw, I've been thinking it could be nice to further enhance the tests using some recorded network traffic with https://github.com/vcr/vcr , I could do that if you don't mind some further delay,  what do you think?

stevecrozz:
I think this is great @crunis. Thanks for persisting! I think your idea of using VCR is a good one. I'm going to merge this one today, but if you want to add some VCR tests, I'm happy to pull those in as well.

PR #1273, opened 2018-05-11
Feature/problems page improvements

burnettk:
there are two main things going on here. first, this fixes a bug i introduced with https://github.com/errbit/errbit/pull/1268, where when you're on the /problems page and you filter by some keyword, the "APP" column disappears. second, this adds a feature that makes errbit usable to people like us who have apps that are very noisy (sending multiple problems, multiple times per minute), such that the problems page UI (above the fold) is entirely consumed with the same problems at all times, and other (more important, possibly new) problems cannot be viewed from this interface (without paging). so this adds an easter egg feature (UI could come later if someone really really wants it) where you can hit urls like `/problems?exclude_apps=noisy_app,another_noisy_app`. totes lit fam?


rud:
A plead for easy review/merge: one focus per PR, like you do with commits. I mean, it's a good bugfix in 0e83d18c02 - it makes good sense to merge without further discussion :)

As for the exclusion filters, it also seems like a neat feature. Maybe add at least a bit of documentation as to how it can be used?

burnettk:
@rud yes, that makes sense, thank you. i will follow this advice in the future. i have added some documentation for the app exclusion filter.

app/models/problem.rb
```diff
@@ -75,6 +74,15 @@ def self.in_env(env)
     env.present? ? where(environment: env) : scoped
   end
 
+  def self.with_app_exclusions(exclude_apps)
+    app_names_to_exclude = exclude_apps && exclude_apps.split(',')
+    if app_names_to_exclude.is_a?(Array) && app_names_to_exclude.any?
```
rud: Suggestion for a simpler implementation if you want it:

``` ruby
app_names_to_exclude = exclude_apps && exclude_apps.split(',')
if app_names_to_exclude.present?
```

burnettk: implemented this suggestion!

burnettk:
i used my extreme git skills to screw up this branch. see replacement pull requests at https://github.com/errbit/errbit/pull/1353 and https://github.com/errbit/errbit/pull/1354. thanks!

PR #1275, opened 2018-06-12
[DOCKER] Remove requirement for having errbit user

mikk150:
Remove all requirement for users in container because you don't really need those and when running containers in not privileged mode(which openshift is running them as) then it says that it cannot update `/app/tmp` directory because we are not allowed to change files/directories which are owned by errbit user

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,24 +1,13 @@
 FROM ruby:2.3.3-alpine
 LABEL maintainer="David Papp <david@ghostmonitor.com>"
 
-ARG UID=101
-ARG GID=101
-
-RUN addgroup -g $GID -S errbit \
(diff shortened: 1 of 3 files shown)


mikk150:
@rud actually this .gitlab-ci.yml file is not needed but may come handy if someone plans to make mirror in gitlab and make gemnasium work again

if you mirror this repo directly to gitlab then this CI config will immidiately start working and publish images to gitlab registry

I just used it because I am lazy and wanted to test this image in my cloud

stevecrozz:
Thanks @mikk150. Can you remove the .gitlab-ci.yml file from the PR? I'm good with merging this without the gitlab file.

mikk150:
@stevecrozz removed .gitlab-ci yml file

stevecrozz:
Looks good. Thanks again @mikk150 !

PR #1276, opened 2018-06-15
avoid 500 error when backtrace object is nil

burnettk:
we got this exception:

````
NoMethodError: undefined method `[]' for nil:NilClass
````

with this backtrace:

````
[PROJECT_ROOT]/decorators/backtrace_line_decorator.rb:18→ file
--
[PROJECT_ROOT]/decorators/backtrace_line_decorator.rb:5→ in_app?
[PROJECT_ROOT]/views/mailer/err_notification.html.haml:31→ block in _app_views_mailer_err_notification_html_haml___2324961091592232048_70027270895640
[PROJECT_ROOT]/views/mailer/err_notification.html.haml:30→ each
[PROJECT_ROOT]/views/mailer/err_notification.html.haml:30→ _app_views_mailer_err_notification_html_haml___2324961091592232048_70027270895640
actionview-4.2.9/lib/action_view/template.rb:145→ block in render
activesupport-4.2.9/lib/active_support/notifications.rb:166→ instrument
actionview-4.2.9/lib/action_view/template.rb:333→ instrument
actionview-4.2.9/lib/action_view/template.rb:143→ render
````

`BacktraceLineDecorator#to_s` already protected against a nil object, so i'm following that pattern with this change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,4 +19,5 @@ bundle
 coverage
 *#
 .ruby-version
+.tool-versions
 NOTES
--- a/app/decorators/backtrace_line_decorator.rb
+++ b/app/decorators/backtrace_line_decorator.rb
@@ -15,7 +15,7 @@ class BacktraceLineDecorator < Draper::Decorator
   end
 
   def file
-    object[:file].to_s
+    object.try(:[], :file).to_s
   end
 
   def method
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by stevecrozz** [comment]
This seems fine to me @burnettk. Can you explain what your notice looks like? The backtrace lines is an array of lines, so it makes me wonder how you end up having a line that's nil. Do you have a backtrace that looks something like?
```ruby
[
  "line1:44",
  nil,
  "line3:55"
]
```

burnettk:
gack, sorry, i already deleted the notice.

stevecrozz:
That's ok @burnettk. I think I only hesitate because it seems like this indicates another problem that we could be masking. Does it ever make sense to have a nil line in a backtrace? At first glance it seems like that's already a problem and we should take care of it when we're inserting notices. But we don't know what happened, and this does seem like a practical solution so I'm going to merge it.

Issue #1253, opened 2018-01-06
Remove support for outdated versions of mongodb


stevecrozz:
Update readme files, docker-compose, travis config and anything else to point to currently supported versions of mongodb. Apprently, there is at least one reason we can't yet support v3.6 as discovered by @rud in https://github.com/errbit/errbit/pull/1250#issuecomment-355765369. Relevant links below:

https://www.mongodb.com/support-policy
https://docs.mongodb.com/v3.2/reference/operator/update/pushAll/#up._S_pushAll

rud:
So, turns out the `mongoid` gem does not support mongodb 3.6-series yet. One of the issues is this: https://jira.mongodb.org/projects/MONGOID/issues/MONGOID-4499

Version 3.4 of mongodb seems reasonable maximum to aim for for now.

PR #1278, opened 2018-07-04
Upgrade Mongoid to 5.4 to support MongoDB 3.6

noniq:
As mentioned in https://github.com/errbit/errbit/issues/1253#issuecomment-358295746, older versions of Mongoid do not support MongoDB 3.6 because they still use `$pushAll`. https://jira.mongodb.org/projects/MONGOID/issues/MONGOID-4499 mentions that this has been fixed for Mongoid 6.3 and 7.0. However, Mongoid 6+ depends on Rails 5+, so this is not a simple upgrade.

Fortunately, the fixes also got backported to 5.4 a bit later, see https://jira.mongodb.org/browse/MONGOID-4538

So if I got this right, upgrading Mongoid to 5.4 should allow using MongoDB 3.6. Let’s see how it goes in CI.


stevecrozz:
Looks good to me @noniq. Would you make the README.md change as well to specify that Mongo 3.6 is now supported?

rud:
That is some excellent spelunking @noniq, good find!

noniq:
@stevecrozz Sure, done!

stevecrozz:
Thanks very much @noniq

PR #1293, opened 2018-07-16
Upgrade jquery-rails to v4

stevecrozz:


stevecrozz:
Finally attempting to upgrade jquery rails. There were some comments hanging around warning that we cannot do this, but I've done it anyway. I'm pretty sure we're going to find some gaps in test coverage with this one. The tests are running green, but I have a hard time believing every little thing is still working after upgrading 2 major versions of jquery-rails.

Also, I decided to dump jquery-alerts rather than make it work. It seems unnecessary to me as the native facilities work great. Anyone have a contrary opinion on this?

rud:
Looks awesome

stevecrozz:
I can't shake the feeling that this is going to break some stuff. I wish I didn't have to leave it to our users to find the problems, but I don't have time to dig in deep. Hopefully anyone who needs real stability is using a released version of Errbit and not master, and the rest will hopefully report issues so we can fill in the test coverage gaps.

Issue #1230, opened 2017-10-12
Deployment on now.sh


ProGM:
Dear maintainers,
I've been using errbit for a while in my projects. Recently I discovered a new deployment platform, [`now`](http://now.sh).

It can easily deploy dockerized projects and it's free for small environments.

I've written a guide to deploy there:

https://github.com/ProGM/errbit/blob/deploy-on-now/docs/deployment/now.md

Would you accept a Pull request?

I have a small concern about my guide:
Since it's impossible to run sh commands in the `now` environment, running `rake errbit:bootstrap` is a big problem.
I've proposed two solutions, but I don't like them very much. If you have better solutions, I'm open to any suggestion.

Thank you

stevecrozz:
I'm not too keen on maintaining support / documentation for new strategies, but I wouldn't mind linking to yours if you plan to maintain it. We could include a link on the deployment strategies page if you'd like to add a pull request: https://github.com/errbit/errbit/blob/master/docs/deployment.md

I would just want to make sure its clear that this strategy is not maintained by errbit. Eventually, I'd like to remove capistrano and heroku as supported strategies also.

rud:
Please feel free to open a pull-request @ProGM if you want to maintain the installation guide going forward. Thanks!

PR #1302, opened 2018-07-17
Remove capistrano as a supported deploy mechanism

rud:
As mentioned in https://github.com/errbit/errbit/issues/1230#issuecomment-355782125 the capistrano deployment mechanism is going away in the future. 

This PR does the work of actually removing the dependency, documentation, and unsupported sample config.


stevecrozz:
I was just thinking about this same thing @rud, prompted by the new gem update nag bot which is probably the same reason you're doing this. I know people did use capistrano in the past. I guess nothing stops the same people from continuing to use capistrano even when we remove it from the Gemfile.

Even so, maybe we should leave this open for a week or so and make sure we don't get a lot of objections.

rud:
If this is merged, fixes #1301

rud:
Rebased on master

rud:
Alright, it's a week later, the build is fully green after a rebase, the code will stay around in the git history. I'm pushing the merge-button. 🚀

PR #1304, opened 2018-07-17
Implement /health/readiness without using Timeout.timeout

rud:
For more on why this pattern is problematic, read https://github.com/ankane/the-ultimate-guide-to-ruby-timeouts and https://www.mikeperham.com/2015/05/08/timeout-rubys-most-dangerous-api/

### DIFF
--- a/app/controllers/health_controller.rb
+++ b/app/controllers/health_controller.rb
@@ -27,7 +27,7 @@ private
     impatient_mongoid_client.collections
     { check_name: 'mongo', ok: true }
   rescue StandardError => e
-    { check_name: 'mongo', ok: false, error_details: "#{e.class}" }
+    { check_name: 'mongo', ok: false, error_details: e.class.to_s }
   end
 
   def impatient_mongoid_client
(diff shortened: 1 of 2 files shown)


stevecrozz:
I agree about Timeout. But there was a purpose here. I think the idea is that if mongo is not ready, we should return some readiness information (not ready) rather than wait for mongo. Is there something else we can do? I suppose we could use Socket to test the connection, but it would be cool if mongo/mongoid had something for us.

stevecrozz:
As a quick test, this seemed to do the trick for me:
```ruby
Mongo::Client.new(Errbit::Config.mongo_url, server_selection_timeout: 0.5, connect_timeout: 0.5, socket_timeout: 0.5).collections
```

rud:
Good point with the faster timeout for this case. 

Stykle suggestion: maybe extract the "impatient" connection config into a private method revealing the intent? Then it's off to the races

stevecrozz:
Yeah, that seems like a good idea. I'm going to leave this open to remind myself to make this change.

rud:
Used your code, extracted a `impatient_mongoid_client` private helper method. Think this is good to go now.

stevecrozz:
Looks great to me @rud. I like the name of the new client too :)

PR #1315, opened 2018-07-23
Sparklines with ajax and rubocop clean

burnettk:
This is mostly work from @danielbeardsley in https://github.com/errbit/errbit/pull/864. I made the sparklines queries happen in a separate ajax call so there is no chance it can slow down the rest of the problems page. I also updated rubocop stuff to hopefully fix the build. Maybe these changes make this ready for merge?

### DIFF
--- a/app/controllers/problems_controller.rb
+++ b/app/controllers/problems_controller.rb
@@ -1,3 +1,5 @@
+require 'sparklines'
+
 class ProblemsController < ApplicationController
   include ProblemsSearcher
 
--- a/app/controllers/problems_controller.rb
+++ b/app/controllers/problems_controller.rb
@@ -46,6 +46,10 @@ class ProblemsController < ApplicationController
     @comment = Comment.new
   end
 
+  def xhr_sparkline
+    render partial: 'problems/sparkline', layout: false
+  end
+
   def close_issue
     issue = Issue.new(problem: problem, user: current_user)
     flash[:error] = issue.errors.full_messages.join(', ') unless issue.close
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -54,32 +54,6 @@ module ApplicationHelper
     collection.to_a[head_size..-1].to_a
   end
 
-  def sparkline(times, start_time, num_buckets = 10)
-    times = times.select { |time| time >= start_time }
-    buckets = bucketize(times, start_time, Time.now.to_i, num_buckets)
-    bars = buckets.map do |val|
-      percent = (val * 100.0).round(2).to_s + "%"
-      "<i style='height:#{percent}'></i>"
-    end.join
-    "<div class='spark'>#{bars}</div>".html_safe
-  end
-
-  def bucketize(values, min, max, num_buckets)
-    get_min = ->(a, b) { a > b ? b : a }
-    range = max - min
-    buckets = Array.new(num_buckets, 0)
-    values.each do |val|
-      normalized = (val - min) / range.to_f
-
-      # Use get_min() here because there will inevitably be one value == max
-      # which means normalized = 1 and bucket[1 * num_buckets] doesn't exist
-      bucket_index = get_min.call((normalized * num_buckets).floor, num_buckets - 1)
-      buckets[bucket_index] += 1
-    end
-    max = buckets.max.to_f
-    buckets.map { |count| count / max }
-  end
-
   def issue_tracker_types
     ErrbitPlugin::Registry.issue_trackers.map do |_, object|
       IssueTrackerTypeDecorator.new(object)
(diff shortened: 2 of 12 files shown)


danielbeardsley:
Excited about this! Thanks for finishing it up!

app/views/problems/_sparkline.html.haml
```diff
@@ -0,0 +1,9 @@
+= now_as_seconds_since_epoch = Time.current.to_i
+= now_as_seconds_since_epoch = 8.weeks.ago.to_i
```
rud: Duplicate definition of variable - think this line 2 needs to go?

burnettk: resolved, thanks for catching this!

stevecrozz:
I know this has been queued up for a while now. I think this can happen. But I want to discuss it a bit before we roll it out. I'm going to put comments in line so we can have a more contextual conversation.

app/controllers/problems_controller.rb
```diff
@@ -46,6 +46,10 @@ def show
     @comment = Comment.new
   end
 
+  def xhr_sparkline
```
stevecrozz: Question. Should this action go in the "API"? Any opinions on that?

app/helpers/application_helper.rb
```diff
@@ -54,6 +54,32 @@ def tail(collection)
     collection.to_a[head_size..-1].to_a
   end
```
stevecrozz: These are fine with me, but let's put them in their own class. ./lib/sparklines perhaps? They seem like static methods, so they can probably be defined as class methods on a `module Sparklines`. Side question: what does this word "sparklines" mean?

app/models/problem.rb
```diff
@@ -207,10 +207,17 @@ def unmerge!
 
     # recache each new problem
     new_problems.each(&:recache)
-
     new_problems
   end
 
+  # Returns an array of unix epoc timestamps for the created_at field of
+  # all notices sincethe since parameter
```
stevecrozz: Does this go fetch all zillion notices, put them in memory and then map them all right here? Or does this syntax correspond to some mongo optimized map-reduce thing? (https://docs.mongodb.com/manual/core/map-reduce/)

I haven't tried it, but if you have a literal million notices, this is likely to be prohibitively slow and tie up puma workers doing a lot of iterating and mapping and such. I *think* there's probably an optimal way to deal with this and still get a usable report from mongo even with millions of notices, but I'm not 100% sure. Any thoughts?

spec/controllers/problems_controller_spec.rb
```diff
@@ -173,6 +173,17 @@
...
   end
 
+  describe "GET /apps/:app_id/problems/:id/xhr_sparkline" do
+    before do
+      sign_in user
+    end
+
+    it "renders without error" do
+      get :xhr_sparkline, app_id: app.id, id: err.problem.id
+      expect(response).to be_success
```
stevecrozz: I think we're going to need something a little more robust in terms of testing here.

app/controllers/problems_controller.rb
burnettk: i like keeping this out of the API so folks don't confuse it for something that they might call from a script or something. it returns html, and it's built just for this rails ajaxy integration. i'm not recommending this necessarily, but one day, if the app were rewritten in reactjs, many of the existing controller actions would return json instead of html, and we could make this one fit that model as well.

app/helpers/application_helper.rb
burnettk: done, also moved more stuff to problems model.

app/models/problem.rb
burnettk: yes, this was not extremely straightforward, but ultimately a good suggestion.  it's about 20 times faster for one data set using mongo aggregations. :)

spec/controllers/problems_controller_spec.rb
burnettk: added a bunch of tests for the underlying data in problems_spec.

stevecrozz:
@burnettk I'm not sure what coveralls is complaining about as coverage actually increases with this PR. I'll go check it out.

stevecrozz:
Yeah, don't worry about the coveralls status here @burnettk. Do you want to add anything else before we merge this?

burnettk:
it's good from my perspective, thanks, @stevecrozz !

rud:
@stevecrozz @burnettk I went ahead with the merge, hope everyone is good with that 

Thank you for your effort @burnettk 💯

stevecrozz:
Yep, all good!

burnettk:
https://github.com/errbit/errbit/pull/1355 is also really hot. :)

burnettk:
this is making us very happy, thanks!

PR #1326, opened 2018-07-26
CI: cache rubocop re-run optimization and mongodb install file

rud:


rud:
This has taken a bit of iterating to get just right, but barring any unforeseen consequences, I think it's all set to go now.

noniq:
Kön. M die a

PR #1336, opened 2018-07-26
Heroku deploy: add demo data too

rud:
Status: user/password is now set correctly. No demo data yet.

### DIFF
--- a/app.json
+++ b/app.json
@@ -36,7 +36,7 @@
     }
   },
   "scripts": {
-    "postdeploy": "env ERRBIT_ADMIN_EMAIL=admin@example.com ERRBIT_ADMIN_PASSWORD=demo-admin bundle exec rake errbit:bootstrap errbit:demo"
+    "postdeploy": "bundle exec rake errbit:bootstrap errbit:demo"
   },
   "addons": [
     "mongolab:sandbox",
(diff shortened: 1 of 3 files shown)


rud:
Closing to have heroku app destroyed so it can be recreated.

rud:
Re-opening to re-deploy

rud:
Think this is ready to test - at least the demo admin user is created correctly. 

@stevecrozz I think you need to look in the Heroku logs to figure out why the `errbit:demo` rake-task doesn't seem to have any effect. That data is not created as far as I can tell.

rud:
Btw, the `heroku_pr_review_app?` method in the seed file takes care of the special case of a Heroku deploy from a PR, unlike a regular deploy using the "Deploy to Heroku" button.

stevecrozz:
Here's what I'm seeing from the build:
screenshot-dashboard heroku com-2018 08 05-17-31-13
[build-log.txt](https://github.com/errbit/errbit/files/2260836/build-log.txt)

stevecrozz:
@rud I also invited you to the heroku team so you can check it out.

rud:
Thank you for the invite. I managed to be logged in with a different account, so you'll unfortunately have to invite me once more. Then we should be good to go :+1:

rud:
I think the log output may be in the application log, the rake tasks are probably run as a one-off worker job after the deploy succeeds

rud:
Merging this as it fixes the login issue for PR apps - demo data is still wip :/

PR #1339, opened 2018-07-30
close mongo connection after creating

burnettk:
without this fix, when errbit upgraded mongoid from 5.0.2 to 5.4.0 (https://github.com/errbit/errbit/pull/1278), our mongo servers started seeing a new connection on every `readiness` request, and for some reason the connections were no longer closed implicitly (not sure how that was happening in 5.0.2). in our context, readiness is run every five seconds by kubernetes, so it was creating many connections pretty quickly, and this ultimately took down mongo servers on a few environments.

### DIFF
--- a/app/controllers/health_controller.rb
+++ b/app/controllers/health_controller.rb
@@ -28,10 +28,12 @@ private
     { check_name: 'mongo', ok: true }
   rescue StandardError => e
     { check_name: 'mongo', ok: false, error_details: e.class.to_s }
+  ensure
+    impatient_mongoid_client.close
   end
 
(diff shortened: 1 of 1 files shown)


stevecrozz:
Thanks @burnettk, I'm sure its related to our change here: https://github.com/errbit/errbit/pull/1304

Good catch!

burnettk:
@stevecrozz , sure thing, and thanks for the quick merge! i will put in a pitch for https://github.com/errbit/errbit/pull/1315 and https://github.com/errbit/errbit/pull/1273 while i'm here. :)

PR #1343, opened 2018-08-06
Chore/emit version details

rud:
This adds the first eight characters of the git version number to the UI (when available). Should make it easier to pin-point what version users are on when reporting issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,10 +67,10 @@ Installation
 Rails applications.
 
 * [Install MongoDB](https://www.mongodb.org/downloads)
-* git clone https://github.com/errbit/errbit.git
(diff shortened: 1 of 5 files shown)


stevecrozz:
Makes sense from a display perspective @rud. But who would be setting this environment variable?

rud:
Heroku sets the `SOURCE_VERSION` variable: https://devcenter.heroku.com/changelog-items/630 - and if it's not set it's simply ignored. I think this strikes a nice balance.

rud:
@stevecrozz taking the liberty of merging this. Should make bug-reports easier to parse going forward.

PR #1354, opened 2018-08-07
add /problems?exclude_apps=app1,app2 feature

burnettk:
...to make problems page usable with noisy apps

this is the other half of #1273 , which i am declining since i did a rebase and added a bunch of commits that i didn't want somehow.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -153,11 +153,13 @@ Metrics/AbcSize:
 # ExcludedMethods: refine
 Metrics/BlockLength:
   Max: 418
+  Exclude:
+  - 'spec/**/*.rb'
 
 # Offense count: 4
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 168
+  Max: 243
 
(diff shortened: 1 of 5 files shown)


stevecrozz:
I do want better filtering and I think this option is useful, but I'd like to consider it as part of a more ambitious filtering scheme. I'm inspired by the filtering options we use right here on github issues. Here's an example of a filter + exclusion search: https://github.com/errbit/errbit/issues?utf8=%E2%9C%93&q=author%3Astevecrozz+-label%3Abug+

@burnettk: Can you think of a way to add the app exclusion filter you want while working toward a future where we have a coherent filtering grammer like github's?

The first thing that comes to my mind is a search like `-app:app1 -app:app2`. Does that seem reasonable?

burnettk:
did it!

stevecrozz:
Seems good @burnettk. Not sure what's going on with Coveralls, but all the new lines you added are covered, so that shouldn't be an issue.

rud:
Cool, merging

PR #1358, opened 2018-08-08
Add note about the possibility of per-app notice grouping

### LINKED ISSUE #1349: Update documentation regarding notice grouping
0.7.0 introduced the feature of configuring notice grouping per app. It would be great to mention that in the documentation, since the [0.7.0](https://errbit.com/docs/v0.7.0/) one still says that notice grouping is global.

stefanchrobot:
Closes #1349.


rud:
Thanks!

PR #1361, opened 2018-08-16
Ensure user_attributes returns a Hash

jboler:
My Airbrake config blacklists all keys: `config.blacklist_keys = [/.*/]` so `user_attributes` returns a String `"[Filtered]"` instead of a Hash. 

When trying to save this String to a `Notice` model, ErrBit throws:

    Mongoid::Errors::InvalidValue: message: Value of type String cannot be written to a field of type Hash
    [PROJECT_ROOT]/models/error_report.rb:65→ new
    [PROJECT_ROOT]/models/error_report.rb:65→ make_notice
    [PROJECT_ROOT]/models/error_report.rb:54→ generate_notice!
    [PROJECT_ROOT]/controllers/api/v3/notices_controller.rb:24→ create

This PR ensures that `user_attributes` always returns a Hash.

### DIFF
--- a/lib/airbrake_api/v3/notice_parser.rb
+++ b/lib/airbrake_api/v3/notice_parser.rb
@@ -69,7 +69,8 @@ module AirbrakeApi
       end
 
       def user_attributes
-        return context['user'] if context['user']
+        user = context['user']
+        return user.is_a?(Hash) ? user : { user: user } if user
 
         {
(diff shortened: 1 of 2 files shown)


rud:
The coveralls service still seems confused.

@jboler a test-case would be an awesome addition, if you can find the time

PR #1368, opened 2018-10-09
Resend notifications when generating notices on resolved problems

jasoncodes:
A bit of history:

* #265 introduced a fix which zeroed out the notice count on a problem when it was resolved. This was done to cause notifications to re-trigger when if it were unresolved when recording a notice.
* This behaviour was changed in 13c8b21f91c07e7970029e87fbc6997bcf1e16a6 to instead reset on unresolve to preserve the notice count on resolved problems.
* This behaviour seems to have been lost in f646cf500aef1d37aeb28a4a11f5bf2a4930837c which reduces the number of queries when recording a notice.

This PR restores the desired resending on unresolving behaviour without requiring the notice count to be changed. This is done by checking if the existing problem is resolved before updating it (unresolving and incrementing notice count) and if so, always triggering notifications.

### DIFF
--- a/app/models/error_report.rb
+++ b/app/models/error_report.rb
@@ -22,6 +22,7 @@ class ErrorReport
   attr_reader :notice
   attr_reader :notifier
   attr_reader :problem
+  attr_reader :problem_was_resolved
   attr_reader :request
   attr_reader :server_environment
   attr_reader :user_attributes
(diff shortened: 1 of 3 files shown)


stevecrozz:
Thanks @jasoncodes. I think this makes a lot of sense and its something I haven't noticed because I never use the resolve feature.

PR #1369, opened 2018-10-09
Update Dockerfile

Rovel:
Update docker image from ruby 2.5.1-alpine to match app.json ruby version

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ruby:2.3.6-alpine
+FROM ruby:2.5.1-alpine
 LABEL maintainer="David Papp <david@ghostmonitor.com>"
 
 WORKDIR /app


stevecrozz:
That coverage report doesn't make any sense so I'm merging this.

Rovel:
nice! thks

r u guys open to a new UI style?

stevecrozz:
Yeah, a new UI would be great. There have been several attempts at this already, but I think they have all been too ambitious. I think in order to make progress we need someone who really wants to lead this effort by organizing a number of smaller, individual projects. Like we could start by building new tables, then new forms, then new dialogs, then new navigation, etc.

Doing it all at once seems to be too much. But I'm happy to talk it over if you are interested.

Rovel:
sure, let me open a issue

PR #1370, opened 2018-10-26
Fix memory leak in readiness health check endpoint

jasoncodes:
Warm up:

* Docker image boots with 116 MiB used.
* `ab -n 2000 -c 10 localhost:8080/health/liveness` raises memory usage up to 250 MiB.
* Running `ab` again does not change memory usage.

Before this PR:

* `ab -n 2000 -c 10 localhost:8080/health/readiness` raises memory usage up to 370 MiB.
* Further runs result in memory usage of 400 MiB, 480 MiB, 600 MiB.

After this PR:

* `ab -n 2000 -c 10 localhost:8080/health/readiness` raises memory usage up to 261 MiB.
* Running `ab` again does not change memory usage.

### DIFF
--- a/app/controllers/health_controller.rb
+++ b/app/controllers/health_controller.rb
@@ -33,7 +33,7 @@ private
   end
 
   def impatient_mongoid_client
-    @impatient_mongoid_client ||= Mongo::Client.new(
+    @@impatient_mongoid_client ||= Mongo::Client.new(
       Errbit::Config.mongo_url,
       server_selection_timeout: 0.5,
       connect_timeout:          0.5,
--- a/app/controllers/health_controller.rb
+++ b/app/controllers/health_controller.rb
@@ -32,6 +32,7 @@ private
     impatient_mongoid_client.close
   end
 
+  # rubocop:disable Style/ClassVars
   def impatient_mongoid_client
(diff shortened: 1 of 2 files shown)


app/controllers/health_controller.rb
```diff
@@ -32,12 +32,14 @@ def run_mongo_check
     impatient_mongoid_client.close
   end
 
+  # rubocop:disable Style/ClassVars
   def impatient_mongoid_client
```
stevecrozz: What do you think about making this a class method instead, to avoid these class variables, for example:
```ruby
def self.impatient_mongoid_client
  @impatient_mongoid_client ||= Mongo::Client.new(....)
end
```

jasoncodes:
> does it still return not ready if mongo suddenly goes away?

Yup. I gave this a good amount of manual testing and it seems to fail and reconnect on timeouts and server disappearing the same as it does without this additional memorisation.

app/controllers/health_controller.rb
jasoncodes: Sure. How does 00cbfcf4 look? Let me know if you want me to squash these fixups into the original commit.

stevecrozz: I think it looks really slick, and also thanks for identifying and fixing this issue!

PR #1396, opened 2018-11-30
allow for viewing a specific notice by id. w/ jasquat

burnettk:
Adds path /notice/:id, which takes you to the specific notice, unlike /locate/:id


burnettk:
should we merge this? thanks.

stevecrozz:
Yes. Thanks @burnettk

PR #1397, opened 2018-11-30
always search by notice id

burnettk:
this PR builds on top of https://github.com/errbit/errbit/pull/1396, so if we want this one without that one (i'm guessing that's unlikely), i can pull them apart. i'm hoping https://github.com/errbit/errbit/pull/1396 is reviewed first.

the only real diff here not in the other PR is this file: https://github.com/errbit/errbit/compare/master...burnettk:feature/always-search-by-notice-id?expand=1#diff-037539c7ba3b471b09f5cd4fe163a69f

it makes it so if you shove a notice id into the top-level search box, it will do what you'd expect, and find the related problem. it's not great in terms of performance or code cleanliness, but it feels useful and one of our users expected it.


stevecrozz:
@burnettk Thank you for asking. I would love to have another maintainer. Do you want to send me an email and we can connect and set up a time to talk about it?

rud (approved):
Seems like a handy shortcut to add

Issue #1385, opened 2018-11-05
Can't merge errors after search


keithlayne:
### Step 1: Are you in the right place?

  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/) tagged with errbit.

### Step 2: Describe your environment

  * Errbit version: 0.8.0-dev (via docker image)
  * OS: linux

### Step 3: Describe the problem:

Merging selected issues after search generates a routing error.

#### Steps to reproduce:

  1. Go to an app page
  2. Use the search field to search for a term
  3. Select some errors via checkboxes
  4. Click 'merge'

#### Observed Results:

  * What happened?  This could be a description, log output, etc.

Generic 404 error page.

#### Expected Results:

  * What did you expect to happen?

Successful merge - this used to work, unfortunately I don't have version details for that.

#### Application Logs:

  * Please include any relevant logs, especially if you encountered an
    application error.

```
Started POST "/problems/search" for 100.96.12.30 at 2018-11-05 18:46:05 +0000

ActionController::RoutingError (No route matches [POST] "/problems/search"):
  actionpack (4.2.10) lib/action_dispatch/middleware/debug_exceptions.rb:21:in `call'
  actionpack (4.2.10) lib/action_dispatch/middleware/show_exceptions.rb:30:in `call'
  railties (4.2.10) lib/rails/rack/logger.rb:38:in `call_app'
  railties (4.2.10) lib/rails/rack/logger.rb:22:in `call'
  request_store (1.1.0) lib/request_store/middleware.rb:8:in `call'
  actionpack (4.2.10) lib/action_dispatch/middleware/request_id.rb:21:in `call'
  rack (1.6.10) lib/rack/methodoverride.rb:22:in `call'
  rack (1.6.10) lib/rack/runtime.rb:18:in `call'
  activesupport (4.2.10) lib/active_support/cache/strategy/local_cache_middleware.rb:28:in `call'
  actionpack (4.2.10) lib/action_dispatch/middleware/static.rb:120:in `call'
  rack (1.6.10) lib/rack/sendfile.rb:113:in `call'
  hoptoad_notifier (2.4.11) lib/hoptoad_notifier/user_informer.rb:12:in `call'
  railties (4.2.10) lib/rails/engine.rb:518:in `call'
  railties (4.2.10) lib/rails/application.rb:165:in `call'
  rack (1.6.10) lib/rack/deflater.rb:35:in `call'
[...]

PR #1411, opened 2019-01-18
#1385: Can't merge errors after search

posiczko:
The issue is related to the fact that the `#problem_table` is emptied
when the search is run, and a new table is inserted in its place with
the result of the search. The existing click binding of the buttons 
at the end ofthe table (merge/delete/etc) is removed when 
the table is removed.

As a fix, `assets/javascripts/errbit.js` introduces global
`bindProblemButtonsActions` function which is extracted from the
body of `errbit.js`, containing handler of for problem buttons click
event. This function is then added to the result of
`views/problems/search.js.haml` to be run on a new problems table
when the search is returned.

Tried to create a rspec test for this, but had problems getting nokogiri 
to compile. Maybe it's time for Rails 6 update? 😅

### DIFF
--- a/app/assets/javascripts/errbit.js
+++ b/app/assets/javascripts/errbit.js
@@ -25,9 +25,7 @@ $(function() {
                              "?copy_attributes_from=" + $(this).val();
     });
 
-    $('input[type=submit][data-action]').on('click', function() {
-      $(this).closest('form').attr('action', $(this).attr('data-action'));
-    });
+    bindProblemButtonsActions();
 
     $('.notice-pagination').each(function() {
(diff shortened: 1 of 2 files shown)


krauselukas:
Hey, just wanted to check the state of this PR, this would fix https://github.com/errbit/errbit/issues/1385 We applied this in our errbit instance and it seems to fix the problem :) Are there any changes needed in this PR or any other things that prevent this from being merged?

stevecrozz:
Thanks for the confirmation @krauselukas . I haven't tested this myself but it looks good to me and if you say it works, I'll merge it.

PR #1417, opened 2019-01-29
Fix: Atom feed backtrace error

zealot128:
Fixes error: NoMethodError: undefined method `number' for #<BSON::Document:0x00000005e11180> Did you mean? numeric?

Not sure, but my production app, as well as a fresh development snapshot does not displays atom feeds. The format of the Backtrace Lines seems to be changed.
This MR fixes that.

### DIFF
--- a/app/views/notices/_atom_entry.atom.haml
+++ b/app/views/notices/_atom_entry.atom.haml
@@ -25,10 +25,10 @@
   - for line in notice.backtrace_lines
     %tr
       %td
-        = "#{line.number}:"
+        = "#{line['number']}:"
         &nbsp;&nbsp;
       %td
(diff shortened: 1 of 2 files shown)


stevecrozz:
Thanks @zealot128. I'm guessing we don't have any tests covering atom feed as this is a pretty old feature. Any chance you want to add a test to ensure we don't break this again in the future?

zealot128:
I've modified the view spec that was already existing. But it was incomplete, as no err/notice objects where provided in the test.

I've created the objects for the test. Only stubbing/building did not work, because the Atom feed wants to generate URLs to the app/problem, and therefore needs persisted ids.

burnettk:
thank you, @zealot128

PR #1435, opened 2019-04-16
Remove capistrano bits

nacengineer:
These files will vary based upon who is deploying this project. They should be managed by the deployer separately from the main errbit git code. These parts will only interfere or complicate a remote fetch and pull of the master project on an end user's deploy of errbit that is utilizing capistrano.


nacengineer:
actually with the USER_GEMFILE option in the Gemfile, which if leveraged mutates the Gemfile.lock file, it might be a good idea to not check-in the Gemfile.lock as It also interfere with a git remote fetch of the errbit code but that would be a separate discussion for your core team.

stevecrozz:
This is true (regarding Capistrano). We don't even document it as a deployment option these days:
https://errbit.com/docs/v0.7.0/deployment.html

I'm inclined to merge this but I'd like to give it a few days to give others a chance to chime in.

nacengineer:
@stevecrozz note you still have a link to capistrano on that page. I did not remove it as a part of this PR

stevecrozz:
You're right. I tried to use a permalink so the info would not change, but the Capistrano instructions are gone from master.

PR #1440, opened 2019-05-24
Remove email Auto-Submitted header

alex-ross:
When for example sending email to a Jira service desk all emails with
this value and header will be ignored becouse it's considered an
auto-reply.


alex-ross:
Still rejected in Jira becouse of bulk, I close this, it's not done.

PR #1446, opened 2019-09-19
Fix webhook to actually send its body

douglaswth:


stevecrozz:
Thanks @douglaswth . That's a pretty surprising bug. My only question is, does this work correctly even though we seem to be not setting the Content-Type header to 'application/json' ?

douglaswth:
Sort of, in that my application probably isn't checking the content-type header at the moment. I can add that as well if you like.

stevecrozz:
Awesome. Thanks for the fix!

stevecrozz:
The rubocops got you

douglaswth:
@stevecrozz Also, do you have access to make a new release of Errbit to DockerHub? It doesn't look like it has been updated in 9 months: https://hub.docker.com/r/errbit/errbit/tags.

stevecrozz:
I made a v0.9.0 release today

floehopper:
Correction:

> I made a ~v0.9.0~ v0.8.0 release today

PR #1448, opened 2019-09-26
Add date filter for problems/search

robinboening:
This is the updated (rebased with current master) PR of https://github.com/errbit/errbit/pull/749

--

Added two input fields with a javascript datepicker to problems/search form to filter problems by date range. first_notice_at and last_notice_at are both considered when searching:

  * Problem which was first noticed at 2014-01-05 will be found if your search criteria is for example 2014-01-01 until 2014-02-01.

  * Problem which was last noticed at 2014-01-05 but first noticed at 2013-12-24 will also be found if your criteria is for example 2014-01-01 until 2014-02-01.

By default the very first notice date of the very first problem is used as the beginning value of the date range. The end value is by default the date of tomorrow. So searching for something but not a date will not lead to wrong results.


rud:
Thank you for the rebase. There are a few failing tests, and a slight drop in test coverage. Getting close :)

robinboening:
Yeah, right. I will fix the issues.

robinboening:
@rud Tests are passing now. I noticed two lines not being covered by tests:

https://coveralls.io/builds/25983231/source?filename=app%2Fcontrollers%2Fproblems_searcher.rb#L34

https://coveralls.io/builds/25983231/source?filename=app%2Fcontrollers%2Fproblems_searcher.rb#L38 

Since those exposed variables are only used in the view context and in view specs they need to be stubbed and hence do not get covered so it needs an acceptance test to cover them. I suggest I add one or two, but it might feel a bit redundant since the rest of the feature is well tested already.

What do you think?

robinboening:
Sometimes talking about enlightens yourself.

> Since those exposed variables are only used in the view context

I just noticed, I should probably move it into helper methods since its not needed in the controller. Then the view specs should be able to access them and they are covered.

Agreed?

robinboening:
Its helper methods now and the view specs are covering them.

rud:
Looks good, haven't tried it out though

stevecrozz:
I tried this feature today and the functionality is there and it works and I think its a good feature as well, but the implementation is a step in a different direction from where I want to go with Errbit search.

What I want to achieve is exactly what GitHub and GMail have implemented for searching issues:
- https://help.github.com/en/articles/searching-issues-and-pull-requests
- https://support.google.com/mail/answer/7190?hl=en

I did find a gist that looks like it could get us started with this kind of search parsing:
https://gist.github.com/robhurring/1477730/f4b78f489f24dec467309e19c4b46da210bbc514

Here's what I got when I tried it:
~~~ruby
irb(main):050:0> s = SearchTerms.new('a:b before:2019-10-02 after:2019-10-01 more terms')
=> #<SearchTerms:0x0000564d61cbc4b0 @query="more terms", @parts={"a"=>"b", "before"=>"2019-10-02", "after"=>"2019-10-01"}, @split=true>
irb(main):051:0> s.parts
=> {"a"=>"b", "before"=>"2019-10-02", "after"=>"2019-10-01"}
irb(main):052:0> s.query
=> "more terms"
~~~

If we want to add some UI to help users control these queries, then we can add those too (preferably using browser-native date picker and maybe under some advanced menu), but I want to get this implemented in the search parser first. I 100% understand if you don't want to dive into this with me, but let me know if you do and we can press on.

robinboening:
Hey @stevecrozz, I see your point and I agree the proposed way is leading to a greater set of nice features. Though, right now I don't have the time to work on this, neither can I tell you when I will find the time. Please leave a message if you start working on this. If I do I will let you know as well.

stevecrozz:
Totally understand @robinboening. I'll keep this open for now.

PR #1457, opened 2019-12-04
Disable sending of performance_stats. Fixes #1453

### LINKED ISSUE #1453: Disable performance monitoring
I upgraded airbake from v8 to v9 and airbrake-ruby from v3 to v4 and noticed some errors in my logs: "Airbrake: unexpected code (404)". Errbit was able to record the error though.

After some investigation it turns out it is due to Airbrake client by [default logs performance stats](https://github.com/airbrake/airbrake-ruby#performance_stats) with `c.performance_stats = true`

So perhaps it should be set to false in the Airbrake [configure instructions](https://github.com/errbit/errbit/blob/d49faa2b405c0912d8bdeab57896ab82eeb5f0f4/app/views/apps/_configuration_instructions.html.haml)?

espen:


stevecrozz:
Thanks @espen

PR #1458, opened 2019-12-10
Update Dockerfile

ugorur:
Fix docker health checks port problem

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,6 +37,6 @@ RUN RAILS_ENV=production bundle exec rake assets:precompile \
 
 EXPOSE 8080
 
-HEALTHCHECK CMD curl --fail "http://$(/bin/hostname -i | /usr/bin/awk '{ print $1 }'):8080/users/sign_in" || exit 1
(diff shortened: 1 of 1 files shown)


stevecrozz:
I was not aware that we could use runtime environment variables in the healthcheck command. Is that how this is working?

stevecrozz:
I guess a followup question I have just to understand your use case a little better is why are you changing the default port for running errbit? I had thought users would have no preference on that since its happening inside a container. Did you pass a custom PORT variable at runtime? If so, why? I had thought that would be used only by non docker users.

ugorur:
For your first question, yes you can use any variable or program in health check if it's in your runing container.

I create a docker swarm service with errbit image and I don't want to expose my errbit url to any other services so I create different network and auto configuration tool for connected networks. 

If any of my service detecte errbil network and givin alias for errbit service, it's add errbit variable automatically to environment variables and create new app for given service to errbit if it's not exist.

I just don't want to add extra 8080 port to variable settings so I started my errbit service with 80 port but it didn't start so I change docker file.

Sorry for the unnecessarily long and bad answer :)

stevecrozz:
Thanks for the explanation! This seems fine to me.

Issue #1463, opened 2020-02-13
[GitHub API] Deprecation notice for authentication via URL query parameters


nitinmohan87:
* Errbit version: 0.8.0
  * OS: Linux

I have few Errbit apps configured using Github authentication. I am getting emails from Github about deprecating authentication using access_tokens via query parameters when I login to my errbit apps using my github login. Here is the email I received
```From: GitHub <noreply@github.com>
To: Nitin Mohan <XXX>
Message-ID: <5e3b383365518_470a3fc9baecd95c8235d@github-lowworker-cd52d7b.cp1-iad.github.net.mail>
Subject: [GitHub API] Deprecation notice for authentication via URL query parameters

Hi @nitinmohan87,

On February 5th, 2020 at 21:48 (UTC) your application (<Github OAuth application name used in Errbit>) used an access token (with the User-Agent Faraday v0.9.2) as part of a query parameter to access an endpoint through the GitHub API:

https://api.github.com/user

Please use the Authorization HTTP header instead as using the `access_token` query parameter is deprecated.

Depending on your API usage, we'll be sending you this email reminder once every 3 days for each token and User-Agent used in API calls made on your behalf.
Just one URL that was accessed with a token and User-Agent combination will be listed in the email reminder, not all.

Visit https://developer.github.com/changes/2019-11-05-deprecated-passwords-and-authorizations-api/#authenticating-using-query-parameters for more information.

Thanks,
The GitHub Team
```
Digging into gems configured, errbit is currently using omniauth-github (v1.3.0). We must update that gem to use [omniauth-github v1.4.0](https://github.com/omniauth/omniauth-github/releases/tag/v1.4.0) which includes a fix for above issue. 

#### Steps to reproduce

  1. Deploy Errbit app configured using Github. Use v0.8.0 version
  2. Login to errbit using github creds

#### Observed Results:

  * Got an email from Github about deprecation notice.

#### Expected Results:

  * Use the recommended way of using headers for authentication.

stevecrozz:
Fixed with #1464

PR #1464, opened 2020-02-14
Upgrade omniauth_github

stevecrozz:
The new version of bundler combined with the current gem dependencies
did not allow for the rails_autolink and mongoid_rails_migrations gems.
Since the last migration was written in 2015 and we barely used
rails_autolink, I removed both gems.


stevecrozz:
Ref #1463

nitinmohan87:
I no longer receive deprecation notice emails when I deployed this change. Thanks @stevecrozz !

PR #1474, opened 2020-05-13
Update docker documentation

pjezusek:
The path to the configuration file in "docker.md" has been fixed. The actual leads to the page with 404 error. Also added "bundle exec" statement before each rake command. Without it, there is an error related with the wrong version of rake.


2called-chaos (approved):

2called-chaos:
Almost anniversary 🙈 Thank you very much for the contribution!

PR #1480, opened 2020-09-03
Public source maps

vladislav-yashin:
This is a proof of comcept for public source maps. Not yet fully test-covered and may have some issues, but already runs in production.

### DIFF
--- a/app/models/cached_source_map.rb
+++ b/app/models/cached_source_map.rb
@@ -8,7 +8,7 @@ class CachedSourceMap
     AirbrakeApi::V3::SourceMapLine.new(source_map, generated_line).original_line
   end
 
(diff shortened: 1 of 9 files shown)


stevecrozz:
This is very cool and it is something that I've wanted in Errbit for a long time. But I am honestly unable to commit to maintaining this and working out all the bugs that will inevitably arise from introducing a new feature of this size. But that doesn't mean we can't bring this in.

@funk-yourself can you commit to reviewing future issues and pull requests that come in related to JavaScript source map support? If you can, then I think we can bring this in and I'll work with you on this PR.

piyushchauhan2011:
Any updates on this?

Issue #1373, opened 2018-10-31
New UI study


Rovel:
I want to suggest new UI elements to errbit, as mentioned by maintainers a breadown by elemnts is preffered

So I suggest these main elements:
- [ ] Nav
- [ ] Buttons
- [ ] Tables
- [ ] Chips
- [ ] Cards

Going to materialize or mdl lite is a tendency, any opinions on using a framework?

stevecrozz:
I don't have a preference on framework. I think the last attempt was using bootstrap. I think if we are going to tackle a redesign. One of the things I'd really want is a more mobile-friendly layout. I'm sure all the latest UI frameworks have great support for mobile-friendly layouts.

Aside from that, I usually like to pick whatever I think is going to be a stable, long-term solution. Things that have only been around for a year or two make me worry that they're going to be gone (or superseded) in another year or two. As far as I know, Errbit has never been redesigned in 8 years. It may be another 8 years before its redesigned again so I think we'd be best served by something that is likely to at least still have a web site with published documentation many years in the future.

If you want to review the last redesign effort, the PR is still open:
https://github.com/errbit/errbit/pull/1063

Rovel:
I like mdl mostly because is plain html, css and js and is mobile first
really lite just put thi in your app layout this you can have really fancy responsive design:
```
/ Material Design Lite
    %script{:src => "https://code.getmdl.io/1.3.0/material.min.js"}
    %link{:href => "https://code.getmdl.io/1.3.0/material.indigo-pink.min.css", :rel => "stylesheet"}/
    / Material Design icon font
    %link{:href => "https://fonts.googleapis.com/icon?family=Material+Icons", :rel => "stylesheet"}/
```
I did all the header and nav yesterday in 2h (30min playing design and colors)
![screen shot 2018-11-01 at 11 14 04](https://user-images.githubusercontent.com/2469500/47857250-f09f4400-ddc7-11e8-9ce5-1d4284d4437e.png)

![screen shot 2018-11-01 at 11 24 35](https://user-images.githubusercontent.com/2469500/47857625-d87bf480-ddc8-11e8-8314-022738317ea3.png)

mdl is now a frozen project, beacuse google like more big and complex things but since is really vanilla based it going to be a LTS in my opinion or you can extract the used classes to your own css and js or fork it

Here is more info about it
https://github.com/google/material-design-lite
https://getmdl.io/customize/index.html

alexanderadam:
[...]

PR #1482, opened 2020-10-06
Fv css test

Rovel:
This a UI change and is related to #1373 
To avoid more dependencies to manage layout for this specific app as @stevecrozz  mentioned i did a remix in dark mode. 
only changing css and som html to fit some stuff inside the divs to work better with the changes
I added some changes to the docker-compose file and added some dev dockerfiles to make easier to develop, but I can cleanup this files if this PR makes sense
Screen Shot 2020-10-06 at 15 00 58


stevecrozz (commented):
Dark mode seems like a cool feature, but doesn't this change force everyone to use it? It seems like using the prefer-color-scheme media query would be a nice way to allow users to opt-in:
https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme

Rovel:
sure I will close this one to make a more proper selector

Issue #1483, opened 2020-10-07
Comment ordering


djmetzle:
* Errbit version: Latest
  * OS: Fedora

#### Step 3: Describe the problem:

We're seeing our comments get displayed out of order. There doesn't seem to be any good explanation for the order chosen.

![image](https://user-images.githubusercontent.com/6876047/95367582-c85cd180-0889-11eb-9aa6-38bd5bfc3db4.png)

Can we sort the comments by timestamp?

#### Steps to reproduce:

Add multiple comments to a problem. Not sure what the exact condition is that causes them to be out of order.

#### Expected Results:

The comments should probably be sorted, preferably by timestamp?

#### Suggested Fix:

https://github.com/errbit/errbit/blob/9f53eaaa0ee966ea2d8015e8eeacf6dde1709601/app/views/problems/show.html.haml#L31

Can we sort there in the view?
```ruby
 - problem.comments.sort_by(&:created_at).each do |comment| 
```

stevecrozz:
I'd suggest putting that logic in the problem decorator, but the idea seems good to me. I'd be happy to look at a PR for this.

djmetzle:
@stevecrozz 
> I'd suggest putting that logic in the problem decorator, but the idea seems good to me. I'd be happy to look at a PR for this.

I spent some time poking around, and honestly, i don't think i'm confident enough with Draper to open a PR using the decorator.

i did find this though, the models for problems and notices both  define `ordered` scopes?
https://github.com/errbit/errbit/blob/9f53eaaa0ee966ea2d8015e8eeacf6dde1709601/app/models/problem.rb#L62
https://github.com/errbit/errbit/blob/9f53eaaa0ee966ea2d8015e8eeacf6dde1709601/app/models/notice.rb#L36

Would it be appropriate to follow that pattern and define a similar `ordered` scope on comments? And use that in the view instead?

stevecrozz:
that works for me

PR #1484, opened 2020-10-07
Order comments in problem views by created_at

djmetzle:
This should fix the bug we're seeing where comments are being rendered
in a scrambled order.

This follows the `ordered` scope pattern from notices and problems:
https://github.com/errbit/errbit/blob/9f53eaaa0ee966ea2d8015e8eeacf6dde1709601/app/models/problem.rb#L62
https://github.com/errbit/errbit/blob/9f53eaaa0ee966ea2d8015e8eeacf6dde1709601/app/models/notice.rb#L36

Closes: https://github.com/errbit/errbit/issues/1483

CC @stevecrozz

### DIFF
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -17,6 +17,8 @@ class Comment
 
   validates :body, presence: true
 
+  scope :ordered, -> { order_by(:created_at.asc) }
+
   def deliver_email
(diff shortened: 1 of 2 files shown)


stevecrozz:
Thanks @djmetzle

PR #1485, opened 2020-10-13
Add dev docker env

Rovel:
Hi, This PR is intended to add a basic development docker files to make easier to build and develop the app.

to use just type:
`docker-compose -f dev-docker-compose.yml build errbit`
then
`docker-compose -f dev-docker-compose.yml up`

I would love to have the label: `hacktoberfest-accepted` to this PR.

### DIFF
--- a/dev-docker-compose.yml
+++ b/dev-docker-compose.yml
@@ -4,6 +4,7 @@ services:
   errbit:
     build:
       context: .
+      dockerfile: dev.Dockerfile
     ports:
       - "8080:8080"
     depends_on:
--- a/dev-docker-compose.yml
+++ b/dev-docker-compose.yml
@@ -2,15 +2,18 @@ version: "3"
 
 services:
   errbit:
-    image: errbit/errbit:latest
+    build:
+      context: .
     ports:
       - "8080:8080"
     depends_on:
       - mongo
     environment:
       - PORT=8080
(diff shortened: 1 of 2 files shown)


stevecrozz:
@Rovel this looks good, sorry for being so late and missing the hacktoberfest deadline. This does look useful. If you are still interesting in contributing, here's the things I think this PR still needs:

- Q: Do we really need a second Dockerfile? It's almost the same as the regular Dockerfile. It seems like we can use the regular Dockerfile and avoid having to add a second dev-only Dockerfile
- Need to add instructions to the README.md file

All in all, it looks nice!

Rovel:
Hi thanks for the review, no problem about the hacktoberfest too.

The Dockerfile  is needed to avoid compiling assets since we read them straight from volume mounted, so to make easier instead of building the image every time some asset change. this makes the dev env image different without asset compile stuff.

I was reading the compose files docs and they have a very nice page about the multifile compose thing
https://docs.docker.com/compose/extends/#understanding-multiple-compose-files
https://docs.docker.com/compose/extends/#different-environments

After reading those I think I can make this docker-compose multi file better just overriding the first one with a couple of changes. Im not sure if the Dokcerfile itself support overrides

Let me know if makes sense keeping the Dev.Dockefile

PR #1493, opened 2021-03-11
disable remote config

espen:
A new phone home feature in Airbrake introduced here: https://github.com/airbrake/airbrake-js/commit/78387b647b7fa2cd89b738171a1999ac29c989eb

This request will fail, I assume because with Errbit you self-host and do not have any project info at Airbrake backend.

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,6 +249,7 @@ var airbrake = new airbrakeJs.Client({
   projectId: 'ERRBIT API KEY',
   projectKey: 'ERRBIT API KEY (again)',
   reporter: 'xhr',
+  remoteConfig: false,
   host: 'https://myerrbit.com'
 });
 ```


2called-chaos:
As far as I understood one should also disable this in the ruby airbrake configuration, yes?

Maybe we should note somewhere that one should disable airbrake functionality that errbit cannot utilize, for example we currently have the following things disabled (I'm not 100% sure if they are all enabled by default)

```
# disable features errbit cannot handle (and spams log)
config.job_stats           = false
config.query_stats         = false
config.performance_stats   = false
config.remote_config       = false
```

I'm not sure if there are even more now, haven't really kept up with airbrake changes

2called-chaos:
As far as I understood one should also disable this in the ruby airbrake configuration, yes?

Maybe we should note somewhere that one should disable airbrake functionality that errbit cannot utilize, for example we currently have the following things disabled (I'm not 100% sure if they are all enabled by default)

```
# disable features errbit cannot handle (and spams log)
config.job_stats           = false
config.query_stats         = false
config.performance_stats   = false
config.remote_config       = false
```

I'm not sure if there are even more now, haven't really kept up with airbrake changes

espen:
I am not 100% sure but I think these are two different things. One is for the gem the other is for the JS reporter.

Issue #1491, opened 2021-03-08
Exceptions that have no message cannot be clicked


2called-chaos:
* Errbit version: master
  * OS: Ubuntu 18.04

### Describe the problem:

Exceptions that have no message are unclickable in the listing. I guess it's not really good practice to have no exception message but Ruby can throw them (my issue here was `Interrupt` which has no message)

#### Steps to reproduce:

  1. Track exception with no message

#### Observed Results:

  * Browser console required to get the link and find out what happened

#### Expected Results:

  * Some clickable text in the where & what column

It's an easy fix but I wanted to ask first if there is any preference on how to solve this (with parentheses or a model method or something)

Proposed change: https://github.com/2called-chaos/errbit/commit/4c0a10421cdbd2b99c609c5e5b50d09f792ef157

stevecrozz:
I'm fine with this. Maybe slightly cleaner would be to add a `link_text` method to the problem decorator and use that. I guess the only decision to make there is whether to use the class text or some kind of non message text like "No message" or something along those lines. I actually don't currently use errbit, so I'll leave that decision to you. Do you want to try implementing this via the problem decorator?

2called-chaos:
Something like this? #1495 

Attempting to decorate "higher" (app_decorate.problems) gives me weird issues and I'm not that familiar with Draper. As for the message I think the error class is more useful if you ever have more than one and one would assume that the error class has a descriptive name if it has no message, right?

I thought about wrapping it in an `<em>` to make it visually different but it has styling tied to it and I guess `<i>` is kinda deprecated.

2called-chaos:
fixed by #1495

PR #1495, opened 2021-03-22
Show error class in problem table if there is no message

2called-chaos:
See #1491


stevecrozz (approved):

2called-chaos:
Err, lemme fix that first

stevecrozz:
This is fine. Thanks @2called-chaos !

Issue #1494, opened 2021-03-17
Libv8 error when deploying errbit to heroku


AdrianNowik:
#### Steps to reproduce:

  1. Follow the official notes how to deploy errbit on the Heroku:
  https://github.com/errbit/errbit/blob/master/docs/deployment/heroku.md
  2. Use the `Deploy to Heroku` button provided in the guide:
  https://heroku.com/deploy?template=https://github.com/errbit/errbit/tree/master
  3. Fill in all required fields in the `Create App` Heroku form.
  4. Change GEMFILE_RUBY_VERSION to 2.7.2, since 2.5.1 (which is entered by default, is not supported on heroku-20)
  5. Run `Deploy App`

#### Observed Results:
I've got an error during the bundle, I won't paste the whole stack trace because it's thousands of lines, so I will only paste bundle output and where it fails.

  ```
  -----> Building on the Heroku-20 stack
-----> Determining which buildpack to use for this app
-----> Ruby app detected
-----> Installing bundler 2.2.11
-----> Removing BUNDLED WITH version in the Gemfile.lock
-----> Compiling Ruby/Rails
-----> Using Ruby version: ruby-2.7.2
-----> Installing dependencies using bundler 2.2.11
       Running: BUNDLE_WITHOUT='development' BUNDLE_PATH=vendor/bundle BUNDLE_BIN=vendor/bundle/bin BUNDLE_DEPLOYMENT=1 bundle install -j4
       Fetching gem metadata from https://rubygems.org/.........
       Fetching rake 13.0.1
       Installing rake 13.0.1
       Fetching concurrent-ruby 1.0.5
       Fetching minitest 5.14.0
       Fetching thread_safe 0.3.6
       Fetching builder 3.2.3
       Installing thread_safe 0.3.6
       Installing builder 3.2.3
       Installing minitest 5.14.0
       Installing concurrent-ruby 1.0.5
       Fetching erubis 2.7.0
       Fetching mini_portile2 2.5.0
       Installing mini_portile2 2.5.0
       Installing erubis 2.7.0
       Fetching racc 1.5.2
       Fetching crass 1.0.6
       Installing racc 1.5.2 with native extensions
       Installing crass 1.0.6
       Fetching rack 1.6.13
       Fetching mini_mime 1.0.1
       Installing mini_mime 1.0.1
       Installing rack 1.6.13
       Fetching addressable 2.4.0
       Installing addressable 2.4.0
       Fetching htmlentities 4.3.4
[...]

Issue #1492, opened 2021-03-10
errbit bundle fails in Arch Linux


vahem2lu:
Errbit version: master 
OS: Arch Linux, Ruby 2.5.0 and Ruby 2.7

### Step 3: Describe the problem:
bundle install fails with error
```
  LINK(target) /srv/errbit/git-errbit/.bundle/ruby/2.5.0/gems/libv8-3.16.14.19/vendor/v8/out/x64.release/mksnapshot
ACTION tools_gyp_v8_gyp_v8_snapshot_target_run_mksnapshot
/srv/errbit/git-errbit/.bundle/ruby/2.5.0/gems/libv8-3.16.14.19/vendor/v8/out/x64.release/obj.target/v8_snapshot/geni/snapshot.cc
make[1]: *** [tools/gyp/v8_snapshot.target.x64.mk:13:
/srv/errbit/git-errbit/.bundle/ruby/2.5.0/gems/libv8-3.16.14.19/vendor/v8/out/x64.release/obj.target/v8_snapshot/geni/snapshot.cc] Segmentation fault
(core dumped)
make[1]: Leaving directory '/srv/errbit/git-errbit/.bundle/ruby/2.5.0/gems/libv8-3.16.14.19/vendor/v8/out'
make: *** [Makefile:195: x64.release] Error 2
/srv/errbit/git-errbit/.bundle/ruby/2.5.0/gems/libv8-3.16.14.19/ext/libv8/location.rb:36:in `block in verify_installation!': libv8 did not install
properly, expected binary v8 archive
'/srv/errbit/git-errbit/.bundle/ruby/2.5.0/gems/libv8-3.16.14.19/vendor/v8/out/x64.release/obj.target/tools/gyp/libv8_snapshot.a'to exist, but it was
not found (Libv8::Location::Vendor::ArchiveNotFound)
        from /srv/errbit/git-errbit/.bundle/ruby/2.5.0/gems/libv8-3.16.14.19/ext/libv8/location.rb:35:in `each'
        from /srv/errbit/git-errbit/.bundle/ruby/2.5.0/gems/libv8-3.16.14.19/ext/libv8/location.rb:35:in `verify_installation!'
        from /srv/errbit/git-errbit/.bundle/ruby/2.5.0/gems/libv8-3.16.14.19/ext/libv8/location.rb:26:in `install!'
        from extconf.rb:7:in `<main>'

extconf failed, exit code 1

[...]

PR #1496, opened 2021-03-23
[WiP] Replace therubyracer with mini_racer (updates v8)

2called-chaos:
Please don't merge yet.

Replacing therubyracer (dead since 2017) with [mini_racer](https://github.com/rubyjs/mini_racer) allows the ancient version of libv8 to get an update from 3.16 to 8.4 which fixes build errors on current systems (#1492 #1494).

- [x] Tests pass (CI fails because it uses bundler 2.2)
- [x] `rails assets:precompile` passes
- [ ] `bundle install` passes with bundler >= 2.2

### Problem with bundler 2.2

As I understand it this is caused by a [bug in libv8](https://github.com/rubyjs/libv8/issues/310) and by a change in bundler which had [unintended side effects](https://github.com/rubygems/rubygems/pull/4127). With how this PR is right now this will cause bundler 2.2 to fail with an error as described in the v8 bug.

As I understand the linked bundler PR should make it so that a lock generated with <2.2 should fallback to old behaviour but I fail to understand what that effectively means as it still fails.

Anyway, for bundler 2.2 we would now need to specify the specific platforms and I'm not sure what to add here. To be able to bundle on my Mac and my Ubuntu machine I need to add the following two:

```sh
bundle lock --add-platform x86_64-darwin-19
bundle lock --add-platform x86_64-linux
```

As I'm not quite sure how backwards compatible this "specific platform" change is, I haven't yet added it to this PR. I just know that darwin alone has 16-19 potentially.

**Maybe we can collect a few common platforms? To get yours: **

```sh
ruby -e 'puts Gem::Platform.local'

# or if that outputs nothing ruby is probably aliased (bundle exec)

\ruby -e 'puts Gem::Platform.local'
```

Let me know what you think.

---

### Update

What irks me a bit is that we don't do any of that in another project which is bundled on 2.2. It still only has the generic "ruby" platform and no darwin/linux specific libv8 versions. When I remove java as a platform from errbit's lock file (so only `ruby` remains) I can bundle without problems on both machines and no platform specific gem versions will be added to the lock file. Unfortunately I have no clue about jruby and Java and why this happens.

### Update 2

There is just something wrong somewhere in libv8 and/bundler. Remember when I said we don't have this problem in another repository? Well...

```sh
$ ruby -v
bundler ruby 2.7.2p137 (2020-10-01 revision 5445e04352) [x86_64-darwin19]

$ bundler -v
Bundler version 2.2.15

$ tail -2 Gemfile.lock
BUNDLED WITH
   2.2.14

$ bundle | grep libv8
Using libv8 8.4.255.0 (x86_64-darwin-19)

$ grep libv8 Gemfile.lock
    libv8 (8.4.255.0)
      libv8 (~> 8.4.255)

$ echo "gem 'dle'" >> Gemfile

$ bundle
[..]
Installing libv8 8.4.255.0 with native extensions
Installing dle 1.0.1
Gem::Ext::BuildError: ERROR: Failed to build gem native extension.
[..]
An error occurred while installing libv8 (8.4.255.0), and Bundler cannot continue.
Make sure that `gem install libv8 -v '8.4.255.0' --source 'https://www.rubygems.org/'` succeeds before bundling.

In Gemfile:
  mini_racer was resolved to 0.3.1, which depends on
    libv8
```

Almost uncut output: https://gist.github.com/2called-chaos/6025fb23a6cc85bb663e7fe62d7d589c

Bundle is fine and functioning, I add a totally unrelated gem and it fails? 😕 Also downloads 1GB every time you try it 🙄

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ rvm:
 env:
   global:
     - COVERAGE=true JRUBY_OPTS=--debug
+    - BUNDLER_VERSION=2.1.4
   matrix:
     - MONGODB=ubuntu1604-4.2.3
     - MONGODB=ubuntu1604-4.0.16
@@ -22,6 +23,7 @@ cache:
 
 before_install:
   - gem update --system
+  - gem install bundler --version 2.1.4
 
 before_script:
   - wget --directory-prefix=mongodb-install/ https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-${MONGODB}.tgz
--- a/Gemfile
+++ b/Gemfile
@@ -86,7 +86,7 @@ group :heroku, :production do
 end
 
 group :no_docker, :test, :development do
-  gem 'therubyracer', platform: :ruby # C Ruby (MRI) or Rubinius, but NOT Windows
+  gem 'mini_racer', platform: :ruby # C Ruby (MRI) or Rubinius, but NOT Windows
 end
 
 gem 'puma'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -171,7 +171,7 @@ GEM
     launchy (2.4.3-java)
       addressable (~> 2.3)
       spoon (~> 0.0.1)
-    libv8 (3.16.14.19)
+    libv8 (8.4.255.0)
     loofah (2.4.0)
       crass (~> 1.0.2)
       nokogiri (>= 1.5.9)
@@ -187,6 +187,8 @@ GEM
     mime-types-data (3.2016.0521)
     mini_mime (1.0.1)
     mini_portile2 (2.5.0)
+    mini_racer (0.3.1)
+      libv8 (~> 8.4.255)
     minitest (5.14.0)
     mongo (2.11.3)
       bson (>= 4.4.2, < 5.0.0)
@@ -301,7 +303,6 @@ GEM
     rb-fsevent (0.10.3)
     rb-inotify (0.9.10)
       ffi (>= 0.5.0, < 2)
-    ref (2.0.0)
     regexp_parser (1.3.0)
     request_store (1.1.0)
     responders (2.4.1)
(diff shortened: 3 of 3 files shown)


2called-chaos:
I came to [the same conclusion](https://github.com/rubyjs/libv8/issues/310#issuecomment-744368475) that due to [Bundler's new enhanced multi platform support](https://bundler.io/blog/2020/12/09/bundler-v2-2.html#multiplatform-support) we now have to add every platform we want to support to the lockfile. Since they may often change (on linux it seems to be simple on Mac not so much) we could just go with adding the following as a setup/upgrade step? The lockfile isn't really guaranteed to be "git clean" due to UserGemfile anyways.

```sh
# [..]
bundle lock --add-platform $(\ruby -e 'puts Gem::Platform.local')
bundle install
# [..]
```

Any objections?

stevecrozz:
I took a look around and I found discourse doing this:
https://github.com/discourse/discourse/blob/5fe1d1f84f9954d9522b59b23bb0392dd10c1634/Gemfile.lock#L469-L475

Without really studying it, I'd be inclined to follow what other popular projects are doing.

2called-chaos:
Adding the platforms works for me on Mac and Ubuntu [but M1-Macs have a whole different issue](https://github.com/rubyjs/mini_racer/pull/186) and we are probably missing jruby variations. And I don't know how to fix travis. I still think this is mostly an libv8 issue.

As this is a rather unresolved issue with 2.2 should we just pin bundler to 2.1.X for now? Like the Dockerfile does.

stevecrozz:
> As this is a rather unresolved issue with 2.2 should we just pin bundler to 2.1.X for now? Like the Dockerfile does.

Yeah, I like this idea.

stevecrozz:
Are you happy with this @2called-chaos ? It LGTM

2called-chaos:
> Are you happy with this

Well I guess. Should I add something to the readme/install instructions? That when an error with v8 occurs while bundling to either use bundler 2.1 or run the above mentioned command (and maybe ask to post the platform here so we can add it to the lockfile)?

stevecrozz:
It works great for me, it works for you and it passes the tests. I'll merge it. We can keep an eye on the issues here and see if there's any trouble.

PR #1498, opened 2021-03-26
Bump mimemagic to version 0.3.9

javierm:
Mimemagic has [yanked all versions previous to version 0.3.7](https://rubygems.org/gems/mimemagic/versions). Since the Gemfile.lock included version 0.3.2, running `bundle install` failed.

See also:

* [Dependency on mimemagic 0.3.x no longer valid](https://github.com/rails/rails/issues/41750)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -188,7 +188,9 @@ GEM
     mime-types (3.1)
       mime-types-data (~> 3.2015)
     mime-types-data (3.2016.0521)
-    mimemagic (0.3.2)
+    mimemagic (0.3.9)
+      nokogiri (~> 1)
+      rake
     mini_mime (1.0.1)
     mini_portile2 (2.5.0)
     minitest (5.14.0)


2called-chaos:
Correct me if I'm wrong but doesn't 0.3.9 stop shipping the file the fuzz is all about? See [mimemagicrb/mimemagic#dependencies](https://github.com/mimemagicrb/mimemagic#dependencies)

I'm not sure if the Rails team is working on a better solution still. Maybe we should wait at least until [there was a follow up by the rails team](https://github.com/rails/rails/issues/41750#issuecomment-806025441)?

Another question, Errbit depends on mimemagic through Hipchat as we don't have active-storage. Does Hipchat still serve any purpose as the service itself doesn't exist anymore in that form? It first became Stride and then Slack

javierm:
> Correct me if I'm wrong but doesn't 0.3.9 stop shipping the file the fuzz is all about? 

AFAIK that's correct. New versions require installing `shared-mime-info` in order to provide the same functionality. So I guess existing Errbit applications might be affected when they upgrade, although to be honest I don't know about it :thinking:. Still better than getting the error they get right now when running `bundle install` IMHO.

I think the Rails team is working on removing the Mimemagic dependency, but as you mentioned Errbit won't be affected because it depends on Mimemagic through Hipchat and not Active Storage.

2called-chaos:
> Still better than getting the error they get right now when running bundle install IMHO.

True, already got me when trying to bundle for development today. As far as I understand linux has this file more often than not. Mac users need to `brew install shared-mime-info` and Windows users will need to manually resolve it.

I'm not sure about Docker but I can't test it atm

I will merge your changes as to recover an installable (and testable) state for now. I will open up a follow up issue whether we can drop the dependency altogether or need additional install instructions in #1499

2called-chaos (approved):

PR #1504, opened 2021-08-25
👷 Add CircleCI config

dounokouno:
Hi, maintainers.

Thank you for publishing a wonderful product on OSS. I used to use other products for a while, but I'm back. When I came back, I noticed that some of the errbit were out of date.

There are a few things to worry about, but [travis-ci.org](https://travis-ci.org), which is used for CI, ended on June 15, 2021 and announced the transition to [travis-ci.com](https://travis-ci.com), but travis-ci.com is credit-based, I thought it would be difficult to use with OSS, so this time I added a setting to run CI with CircleCI. If you prefer [GitHub Actions](https://github.com/features/actions), I would like to add a CI setting in GitHub Actions.

https://circleci.com

If you want CircleCI to work with your errbit organization, someone with errbit organization must create an account on CircleCI and work with GitHub to run that branch of errbit on CircleCI. If you can merge this PR into master first, I think that you can run test on CircleCI just by linking CircleCI and GitHub.

For the time being, I added a setting to test Ruby 2.5 and 2.6, MongoDB v4 or later in a matrix, and confirmed that the test passes in all environments. See also the screenshot below for the test results.

Screenshot 2021-08-25 10 03 21

If you would like to use CircleCI for your CI, I would like to create a PR that removes the Travis CI badge etc. after merging this PR. After that, I would like to update Ruby and various libraries. (Since [Ruby 2.5 is no longer supported with Ruby 2.5.9 released on April 5, 2021](https://www.ruby-lang.org/en/news/2021/04/05/ruby-2-5-9-released/).)

Please confirm it.

Thank you for reading to the end.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,6 +51,7 @@ jobs:
       - save_caches
       - run: ruby -v
       - run: bundle -v
+      - run: bundle exec rake errbit:bootstrap
       - run: bundle exec rspec
       - run: bundle exec rubocop
 workflows:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,63 @@
+version: 2.1
+commands:
+  restore_caches:
+    steps:
+      - restore_cache:
+          keys:
+            - rubygems-{{ checksum "_rubygems_cache_key" }}-{{ checksum "Gemfile.lock" }}
+            - rubygems-{{ checksum "_rubygems_cache_key" }}-
+            - rubygems-
+  create_cache_key_files:
+    steps:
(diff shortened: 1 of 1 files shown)


dounokouno:
I forgot to execute `bundle exec rake errbit:bootstrap`, so I added it.

stevecrozz (approved):

stevecrozz:
Thanks @dounokouno. I set up the CircleCI project and it seems to be working. The free CircleCI plan only gives us one level of concurrency, so it might be a good idea to cut back on the number of Mongo versions we test and maybe we should even rethink the number of Ruby versions we support as well. Now that we have Docker, I'm not sure we need to support more than one version of Ruby.

PR #1505, opened 2021-08-28
Fix from travisci to circleci

dounokouno:
Hi,

After confirming that CircleCI works fine, I replaced the Travis CI information with the CircleCI information. Incidentally, I changed the year of copyright to this year. If you don't need this change, please comment and I'll remove it.

However, the CircleCI badge is not displayed, so you may need to change some settings on the CircleCI side. If you have authority to the errbit organization, I would appreciate it if you could check it.

Thank you.


dounokouno:
I just checked and found that the CircleCI badge was displayed 👍 

Screenshot_2021-08-28_10_11_30

stevecrozz:
Looks good. Thanks @dounokouno !

PR #1513, opened 2022-03-28
Configure Ruby 2.7 for Heroku

munen:
Based on the doc update that Errbit now requires Ruby 2.7 (https://github.com/errbit/errbit/commit/c4153cf19c1a7751777ea112fdcd01308bbbb484), update the Heroku config for the currently available Ruby 2.7 version on Heroku (https://devcenter.heroku.com/articles/ruby-support#ruby-versions).

As a side benefit, this allows using the current Heroku stack-20.


munen:
Would you mind explaining why you closed this PR? Without this change, there's no way to use the current Heroku stack-20 whilst stack-18 is deprecated. Also, the errbit docs claim to require Ruby 2.7.

I'm a bit confused.

PR #1514, opened 2022-03-28
Document how to disable polling airbrake remote config

munen:
Following the readme, I used the `airbrake` gem and followed the airbrake docs for configuration. I ran `gem install airbrake` and pointed airbrake to my errbit instance.

In the `development` Rails environment, this worked fine. In the `production` Rails environment, this yielded the following error:

```
E, [2022-03-25T10:10:04.695662 #1] ERROR -- : <?xml version="1.0" encoding="UTF-8"?>
<Error><Code>AccessDenied</Code><Message>Access
Denied</Message><RequestId>WD57Y1AJCJ74WRKH</RequestId><HostId>W+P5KTtP4PAuz+U6mxLhJ5bbKzPuOPcvBRQTGezTtWb5oPtEyz8w+Z+WSedYdOYF3aif72tIcEY=</HostId></Error>
```

After quite a bit of searching and reading/patching airbrake code with @branch14, we saw that airbrake supports a feature called "[remote config](https://github.com/airbrake/airbrake-ruby/pull/636)", enabled by default. When using Errbit, there's no need to poll from airbrake.io, so disabling this feature is readily available.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@
 
 Errbit is a tool for collecting and managing errors from other applications.
 It is [Airbrake](http://airbrake.io) API compliant, so if you are already using
-Airbrake, you can just point the `airbrake` gem to your Errbit server.
+Airbrake, you can just point the `airbrake` gem to your Errbit server (see [howto](#configure-airbrake-gem)).
 
 <table>
   <tr>
@@ -234,6 +234,46 @@ it will be displayed under the *User Details* tab:
 ![User details tab](https://errbit.com/images/error_user_information.png)
 
 This tab will be hidden if no user information is available.
+
+Configure Airbrake gem
+--------------------------
+
+## Bundler
+
+Add the Airbrake gem to your Gemfile:
+
+```ruby
+gem 'airbrake'
+```
+
+## Manual
+
+Invoke the following command from your terminal:
+
+```bash
+gem install airbrake
+```
+
+### Configuration
+
+```bash
+rails g airbrake
+```
+
+This command will generate a configuration file under
+`config/initializers/airbrake.rb`.
+
+Airbrake supports polling a remote config from airbrake.io, enabled by default. However, since you are using Errbit and not Airbrake, this request will fail and generate an XML error message on stdout. To disable polling the remote config, add the following flag to `config/initializers/airbrake.rb`.
+
+```ruby
+c.remote_config = false
+```
+
+
+Configuration
+-------------
+
+https://github.com/airbrake/airbrake
 
 Javascript error notifications
 --------------------------------------


2called-chaos:
I'm not sure where I saw this once but we had docs that locked the airbrake version as newer versions only add features that errbit does not support. We use edge airbrake and our config is rather massive just to restore how it was with (I think) version 4 or 5 which had way better defaults (imho).

I gisted our config 4 and 2 years ago (the latter I linked in some issue I'm sure). Here is our current config: https://gist.github.com/2called-chaos/07a62137b75a4bf2261a328bc94ed26f

Note L32-L35 which disables stuff, you probably hammer requests to errbit that it doesn't understand with the stats features.

Not sure if we can craft a good doc from this. What you guys recon most people are interested in from my example?

munen:
Locking the airbrake gem to an older version makes sense. I did contemplate doing that, but I didn't want to make assumptions regarding it. However, it makes sense that the airbrake gem will create all kinds of requests that are out of scope of errbit. Hence, it could make sense to phrase the errbit readme to target a specific older version.

> Do you want to add the other features to the list before merge?

Thank you for providing your config as a snapshot :pray: 

I'm uncertain what is best for the errbit docs. I like the super concise doc on current `main`.  To be honest, I tried to keep the additional text and config to a minimum and already felt kind of guilty to replace "use airbrake and point to errbit" with a 40 LOC change. But I didn't see a more concise way.

After reading your comment, the best option I see would be to lock the airbrake gem to a 'known' old version and add a complete configuration block to it. This way, there would be no requests that errbit doesn't understand plus the doc doesn't erode when airbrake changes config flags or adds new features.

2called-chaos:
Yeah maybe we can get some more input. I found it btw, if you create a project it shows you this:

https://github.com/errbit/errbit/blob/main/app/views/apps/_configuration_instructions.html.haml#L8

I reckon trying to configure these settings whilst the version is too old will raise errors. So maybe the better solution is to make it more clear in the Readme to use v5 or go to edge and document things but as I said the defaults are 😑

burnettk:
thank you very much @munen and @2called-chaos . yes, it looks like there is an existing view (baked into the app as a rails view) at https://github.com/errbit/errbit/blob/c30750c81964a685e20883c7f894f87ccd1bc324/app/views/apps/_configuration_instructions.html.haml, in case you were able to update that as well or merge the two docs. https://gist.github.com/2called-chaos/07a62137b75a4bf2261a328bc94ed26f#file-z_airbrake-rb-L31-L35 is good stuff and might deserve inclusion in addition to the property you mentioned.

munen:
@burnettk Thank you for merging :pray: 

Good catch on the config instructions view! That one, I didn't see, to be honest *cough*. I did look over the documentation, though(; Having said that, I think it's a great place for beginners to show the config, as well. Maybe I was too confident having used Errbit in what feels like eons in all kinds of projects before^^

I'm happy to merge the two points of documentation. Proposal:

- Rephrase the readme to be very simple, again "use the airbrake gem and configure it like this". 'this' links to a new markdown file in the repo which contains the current docs from https://github.com/errbit/errbit/blob/main/app/views/apps/_configuration_instructions.html.haml
- https://github.com/errbit/errbit/blob/main/app/views/apps/_configuration_instructions.html.haml reads same markdown file

This yields the benefit that there's no two places containing docs for the same thing. Additionally, if there's no need for an airbrake gem version > 5, there's no need to document how to disable airbrake-specific features.

What do you think?

burnettk:
that sounds good. i'm using essentially every version of the airbrake client gem in various apps, including version 12 (i haven't apparently upgraded to 13 anywhere, but i'm sure it can be made to work). so my bias would be to recommend using the latest version you can use (based on the ruby version in your app, etc) and including configuration in the documentation that makes the latest airbrake version work (maybe we suggest 12 or 13 right now). if there is ever a version of airbrake that CANNOT work with errbit, ideally we try to fix that upstream. that said, pick the version that makes sense to you for the docs; i'm happy with any doc improvements. :)

Issue #1507, opened 2021-09-12
[FEATURE REQUEST] Allow ignoring errors


00dav00:
In other error handling platforms you have possibility to ignore an error (or group of errors) that you don't necessarily want to stop raising but would prefer not to receiving notifications for.

They idea would be to have something like the `Ignore` button I added in the screenshot.

![Screenshot from 2021-09-12 10-26-43](https://user-images.githubusercontent.com/9206778/132993451-03c07877-a8e9-40b6-a555-a6e71ba3d2fa.png)

After ignoring an error or group of errors, they will continue to appear in your dashboard, but you won't receive notification for them any longer.

How do you feel about adding support for this feature? With some guidance I would be glad to implement this since we do have some cases in our company were this can be `helpful`.

burnettk:
thanks for raising this, @00dav00 , and re-raising it, @2called-chaos . yes, this does seem like a useful feature, especially if you don't have the ability to easily change the client code to filter on that side. perhaps consider the term "Mute" for that button? Would that be a better metaphor? But yes, a pull request would likely be accepted here.

stevecrozz:
Are you potentially looking for a way to control notices without interaction? We have a configuration called ERRBIT_NOTIFY_AT_NOTICES which controls whether you receive notifications for every notice or only on, for example, the 1st, 50th and 500th occurrence:
ERRBIT_NOTIFY_AT_NOTICES=[1,50,500]

goetzc:
I think the main idea here is to be able to silent or ignore a specific error from within the GUI. `ERRBIT_NOTIFY_AT_NOTICES` is nice but affects all errors in a specific project.

stevecrozz:
Can we describe the particular use case a bit more and ? My understanding is that we're describing a situation like:

- ERRBIT_NOTIFY_AT_NOTICES=[0] (notify for every notice)
- User is getting too many notices for a particular error and is getting annoyed
- User doesn't want to change the notice behavior for all notices (can we dig into why not? what is so special about this notice?)
- User can't or doesn't want to change the client to stop sending notices to Errbit
- User wants Errbit to behave normally, but stop all future notifications

A couple more questions:
- If you mute a problem, can it be unmuted?
- Where do we store the mute bit?
- Is it permanent? or a temporary mute?

00dav00:
Hi @stevecrozz , the questions you posted seems quite relevant, let me try to answer them.

[...]

PR #1516, opened 2022-03-28
doc: How to configure Airbrake gem (in the Readme and code)

munen:
As discussed in https://github.com/errbit/errbit/pull/1514, harmonize the Readme and generated documentation.


app/views/apps/_configuration_instructions.html.erb
```diff
@@ -0,0 +1,23 @@
...
+  config.project_key = '<%= api_key %>'
+
+  # Airbrake supports polling a remote config from airbrake.io,
+  # enabled by default. However, since you are using Errbit and not
+  # Airbrake, this request would fail and generate an XML error
+  # message on stdout.
+  config.remote_config = false
+
+  # The same goes for performance_stats
+  config.performance_stats = false
```
burnettk, line 22: what about these two, or are they not needed? https://gist.github.com/2called-chaos/07a62137b75a4bf2261a328bc94ed26f#file-z_airbrake-rb-L32-L33

munen: They are certainly fine. I've realized it at the same time you wrote this. Rebased and pushed.

munen:
@2called-chaos provided a [nice gist](https://gist.github.com/2called-chaos/07a62137b75a4bf2261a328bc94ed26f#file-z_airbrake-rb-L32-L33) with an elaborate config.

I omitted most of the flags, because:

1. I wanted to keep the docs to a minimum.
2. Since we document using the generator, the users will always have up to date flags.
3. Testing the more complicated (and probably as beneficial) flags is outside the scope of what I can do right now. It's probably also better if a core maintainer can take a look at them.

burnettk:
there's a lot of good stuff in the gist in terms of making useful stuff from the request env available, filtering out less useful stuff from the request env, ignoring less useful exceptions, etc. i went through and grabbed some stuff for our env, but i'm not sure it's the sort of thing we can document as "you will definitely want this in your configs" unless we make it a lot smaller. i'd definitely be open to a pared-down version of this that we could add to the new configuration instructions that just includes the uncontroversial bits, in case someone wants to take a stab at that.

2called-chaos:
It's been a long time since I've used airbrake v5 but I remember being rather annoyed with v6 as it reported literally everything and without any useful annotation data like v5 had. To be frank I haven't tried an "empty" config recently, does it still omit env and headers by default? I would value this as pretty important and maybe that is worth including then.

Ignoring exceptions can be important, it certainly is for us as we would get thousands of AR::NotFound alone for example. I suppose this isn't as big of a deal if the backend supports ignoring errors which Errbit does not (#1507) but depending on how such an ignore function would work it might still get spammy as hell if an unnice person deems it necessary to vulnerability scan your site (which raises many errors in Rails' middlewares so you can't even rescue in your app).

But there is a reason I just posted my config from time to time rather than make a doc out of it, I don't know what normal people want or need (I tend to go overboard with options and flexibility). And certainly do I not have a good idea of where to place it. I too want to keep the instructions and readme as concise as possible but maybe we can link to/hint at a doc file in the repo or wiki with "advanced config examples" or something along those lines.

But then again most if not all [is already kind of documented](https://github.com/airbrake/airbrake-ruby#config-options) but from what I gathered from you guys is that nobody really wants to read all this (myself included). While skimming through I just learned that setting `performance_stats` to false makes it unnecessary to disable the other stats separately.

So all in all maybe we just link to the official documentation and maybe provide a (cleaned up) version of my gist as inspiration for what to ctrl-F for :D

burnettk:
i think i'm on the same page with you on pretty much all of this. https://github.com/airbrake/airbrake/blob/master/CHANGELOG.md#v610rc1-may-10-2017 talks about the "omit env and headers by default" thing. i've always wondered where those went, thanks for the lead about it changing in airbrake 6. so we should probably re-shove some stuff from the context back into the env in our example configs somehow.

Issue #1512, opened 2022-02-05
docker image errbit/errbit:latest


nijikon:
Can we update `errbit/errbit:latest` docker image on docker hub (https://hub.docker.com/r/errbit/errbit)? What is the procedure for that? The last update was 10 months ago.

stevecrozz:
We used to get this on every commit with the dockerhub GitHub integration
and trusted builds. Dockerhub no longer provides this public service for
free. They do have an open source account and I have applied on behalf of
errbit, but haven't heard back yet. If anyone at @docker is listening, can
you bump us up the queue a bit? I can push builds but trust-wise seems a
little iffy for folks to be deploying builds I generate with no paper
trail. Maybe there's other options? Does GitHub provide anything for this
use case now?

nijikon:
@stevecrozz let me know if there is anything I can help you with in regards to this issue.

stevecrozz:
Do you want to see if GitHub actions can be used to build and publish
images to dockerhub? It seems like that should work and if you can get it
working I'd be happy to accept a PR

nijikon:
@stevecrozz sounds good, will give it a shot this week.

shreve:
There is now an action in place that will update `:latest` from main, as well as create semver images whenever a new tag is pushed.

https://github.com/errbit/errbit/actions/workflows/publish.yml

stevecrozz:
Thanks @shreve !

xxz3:
I have a problem with `errbit:latest` image from dockerhub. When building image it can't find puma gem:

```
errbit_1  | bundler: command not found: puma
errbit_1  | Install missing gem executables with `bundle install`
errbit_errbit_1 exited with code 127
```

`errbit:v0.9.0` works fine. Do you have any hint on how to build the latest version?

shreve:
@xxz3 I'm not able to reproduce this issue, but whatever is going on its not tied to this issue. `errbit:latest` is now updated based on the main branch automatically, so if there really is a problem, it's a new issue.

PR #1517, opened 2022-03-30
Add publish GitHub action workflow

shreve:
Addresses #1512 

This PR adds a workflow definition file that will build and publish the image to dockerhub when you push a tagged version.

You can see a demo of how this worked on my fork:

```
git commit -m "Add publish GitHub action workflow"
git tag v9.9.0
git push --tags
```

Which triggered [this build](https://github.com/shreve/errbit/runs/5760542942), and pushed the image to [this Docker Hub repo](https://hub.docker.com/repository/docker/shreve/errbit/tags).

As you can see, it creates the layered semver tags: `9`, `9.9`, `9.9.0`, and the ever-useful `latest`.

The only configuration I needed to add was `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN` in the repo actions secrets.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -2,6 +2,8 @@ name: build-and-publish
 
 on:
   push:
+    branches:
+      - main
     tags:
       - '*'
 
@@ -25,6 +27,7 @@ jobs:
         with:
           images: ${{ github.event.repository.owner.name }}/${{ github.event.repository.name }}
           tags: |
+            type=raw,value=latest
             type=semver,pattern={{version}}
             type=semver,pattern={{major}}.{{minor}}
             type=semver,pattern={{major}}
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,44 @@
+name: build-and-publish
+
+on:
+  push:
+    tags:
+      - '*'
+
+jobs:
+  build-and-publish:
+    runs-on: ubuntu-latest
+    steps:
+
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Login to DockerHub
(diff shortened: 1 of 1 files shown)


stevecrozz:
Neat! Can you explain it just a little for me? Specifically, I can't tell what qemu is being used for and what this buildx thing is. Is this about cross-compiling errbit for a different CPU architecture?

burnettk:
nice work. i assume the proposal here would be to keep errbit/errbit on docker hub and not move to shreve/errbit? errbit/errbit seems nicer if we can get creds working like shreve has, but either way works for me.

shreve:
> Neat! Can you explain it just a little for me? Specifically, I can't tell what qemu is being used for and what this buildx thing is. Is this about cross-compiling errbit for a different CPU architecture?

@stevecrozz That's a great question. I started this from a sample in the build-push-action repo, and yes, it seems that QEMU and Buildx aren't needed. I've removed them and it still appears to build correctly.

> nice work. i assume the proposal here would be to keep errbit/errbit on docker hub and not move to shreve/errbit? errbit/errbit seems nicer if we can get creds working like shreve has, but either way works for me.

@burnettk Yes, the line that configures the image name uses some variables from the action that pull the username and reponame, so if merged here, it should go to `errbit/errbit`. It also means forks will automatically build to their own private repos.

```yaml
images: ${{ github.event.repository.owner.name }}/${{ github.event.repository.name }}
```

Whoever controls this repo and the Docker Hub repo just needs to generate an auth token and add it in the actions secrets here.

stevecrozz:
Let's give it a shot. I added some secrets. I guess it should build new images whenever we push a tag?

stevecrozz:
@shreve can you make it so that it pushes a "latest" image to dockerhub on every commit to master?

shreve:
There was an incident earlier today where Github Actions were apparently getting blocked. I think that's related to what's happening with build # 1 here. https://www.githubstatus.com/incidents/vxvyrmy9w1vp

stevecrozz:
Right on @shreve . It looks like this works great. At least we have a new latest image! May have to cut a 0.10.0 soon just to test this out.

PR #1519, opened 2022-05-26
Update README.md

nacengineer:
Add the missing closing tags on the descriptive list in the Readme file

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,10 +112,10 @@ Authentication
 If you host Errbit at errbit.example.com, you would fill in:
 
 <dl>
-<dt>URL
-<dd>http://errbit.example.com
-<dt>Callback URL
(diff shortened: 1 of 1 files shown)


stevecrozz:
This is fine with me, but just out of curiosity, is this causing a problem? I thought that HTML5 allowed for dd tags to be unclosed.

nacengineer:
hmm. Well I'm trying to parse it with [this library](https://github.com/pragdave/earmark) for an internal application that fetches Readme's via Gitlab API. So maybe it's actually an issue with the parser. 🤔

stevecrozz:
Probably, but if its causing you trouble, I'm fine with making this change.

nacengineer:
ty! 👍🏻

PR #1521, opened 2022-05-30
Upgrade rails

biow0lf:
WIP

Upgrade Rails to 7.0.

Complete list of changes:
* Ruby 3.1
* Rails 7.0
* Etc…


stevecrozz:
This is very exciting. I am following along with your progress :)

biow0lf:
@stevecrozz I try and understand that porting rails 4.2 application directly to rails 7.0 is hard and break too much. So, I port it to 5.0 and testing in my private branch.

For now, I have two branches with back ported changes. [First](https://github.com/errbit/errbit/pull/1526) and [second](https://github.com/errbit/errbit/pull/1525).

raddrick:
would be happy to help if i can. lmk

burnettk:
@raddrick yes, please. i don't think there is any active work in this area, so if you jump in, i don't think you'll be in conflict with anyone.

biow0lf:
> @raddrick yes, please. i don't think there is any active work in this area, so if you jump in, i don't think you'll be in conflict with anyone.

Just linking my reply from my PR for rails update.

https://github.com/errbit/errbit/pull/1533#issuecomment-2260699359

PR #1522, opened 2022-06-05
Binstubs fixes and cleanup

biow0lf:
This is part of rails 7 changes that are not for rails 7 itself.

1. Remove useless bundler binstubs. Like: `bin/byebug`, `bin/haml`, etc. Nobody use them while development or production.
2. Fix ruby version in dokku installation documentation.
3. Regenerate `bin/rails` and `bin/rake` via `rake rails:update:bin`. This is fix production rails console:

```console
/app # rails c
Looks like your app's ./bin/rails is a stub that was generated by Bundler.

In Rails 4, your app's bin/ directory contains executables that are versioned
like any other source code, rather than stubs that are generated on demand.
```


biow0lf:
@stevecrozz can you review and merge?

burnettk:
It's possible someone was depending on the binstubs instead of running `bundle exec`. that person is not me, and we could always add them back if anyone wants them. approved. :)

stevecrozz:
I think you're right @biow0lf. I certainly don't use any of those binstubs that were deleted.

PR #1524, opened 2022-06-13
Added support for custom_backtrace_url's.

gamecreature:
This pull request allows the configuration of clickable backtraces to custom repositories.
Errbit currently supports gitbucket and github links via the _repo app fields.
With this change it is possible to define a custom url (with fields) for an app. 

An app has an extra field named: 'custom_backtrace_url'.
This can be used to define a custom link to a gitea and gitlab repository (and possibly other systems).

Examples:
- gitea: https://example.com/repo/name/src/branch/%{branch}/%{file}#L%{line}
- gitlab: https://example.com/repo/name/-/blob/%{branch}/%{file}#L%{line}

The available fields:
- branch: raw branch name 
- file: the filename name 
- line: the line number
- ebranch: escaped branch name (Via CGI.escape)
- efile: escaped file name (via CGI.escape)

### DIFF
--- a/docs/apps.md
+++ b/docs/apps.md
@@ -26,11 +26,10 @@ There is no UI for this feature, just the query param.
 
 ## Linking backtraces to custom repositories
 
-Errbit builds backtraces with clickable links to Github and bitbucket repositories.
+Errbit builds backtraces with clickable links to Github and Bitbucket repositories.
 The entry 'custom backtrace URL template' can be used to support clickable backtraces with other repositories.
-It's possible to build a template URL to link to a certain repository.
 
-You can use the following fields in this template:
+The following fields are available for this template:
 
 - %{branch}: The repo branch name
 - %{file}: The relative file/path name of the backtrace file


app/models/app.rb
```diff
@@ -126,6 +127,15 @@ def bitbucket_repo?
     bitbucket_repo.present?
   end
 
+  def build_custom_backtrace_url(file, line)
+    format(custom_backtrace_url, branch: repo_branch, file: file, line: line,
+                                 ebranch: CGI.escape(repo_branch), efile: CGI.escape(file))
```
stevecrozz, line 132: Can you explain the need for both escaped and unescaped names? It seems surprising that the escaped branch alone is insufficient.

gamecreature: Wel in my test-cases I've only used/needed the unescaped values. I've added the ebranch and efile names, to support other url's. 
CGI.esacpes also escapes the '/' token. Which doesn't work for the gitea and gilab pathnames and branchnames. 

If you want I can remove the ebranch and efile fields. (They can be added later if somebody runs into problems with other git repos)

app/decorators/backtrace_line_decorator.rb
```diff
@@ -64,7 +64,7 @@ def link_to_in_app_source_file(app, text)
   end
 
   def link_to_repo_source_file(app, text)
-    link_to_github(app, text) || link_to_bitbucket(app, text)
+    link_to_github(app, text) || link_to_bitbucket(app, text) || link_to_custom_backtrace_url(app, text)
```
stevecrozz, line 67: What do you think about putting the custom URL first? Seems like if you have one, you'd expect it to be used regardless.

app/models/app.rb
stevecrozz: Seems useful enough to keep. I was just curious.

app/models/app.rb
```diff
@@ -7,6 +7,7 @@ class App
   field :api_key
   field :github_repo
   field :bitbucket_repo
+  field :custom_backtrace_url
```
stevecrozz, line 10: The only remaining thing on my mind is this name. Which because it is a database field, we probably will never be able to change it. I'm trying to decide between this or 'backtrace_url' or 'backtrace_url_template' Have you considered any other names?

gamecreature: My initial thought was custom_repo, but it's not correct. 
perhaps backtrace_url_template is a better term.  

When change the name to backtrace_url_template, IIcan change the method build_backtrace_url to backtrace_url. Which seems better.

stevecrozz: I like that!

biow0lf: Maybe `custom_repo`? Just to be in sync with `github` and `bitbucket` `_repo`.

biow0lf:
@gamecreature can you add examples somewhere in docs?

app/models/app.rb
stevecrozz: I like the inclination toward consistency, but I think in this case these are too different to use the same name. One is a reference to a repo and the other is a template to use for a backtrace link. Maybe someday we can make all the backtrace links work using this new field even for GitHub and Bitbucket.

gamecreature:
@stevecrozz The field has been renamed to `custom_backtrace_url_template`. 
@biow0lf I've just added examples to the app.md docs

stevecrozz:
Thanks @gamecreature !

Issue #838, opened 2015-02-19
Unexpected Behavior with ENV in Gemfile


mtmcfarl:
I recently got a fresh clone of errbit and put it on my dev machine and cd into the directory after which I was greeted with:

``` Unknown ruby interpreter version (do not know how to handle): ENV[GEMFILE_RUBY_VERSION]ifENV[GEMFILE_RUBY_VERSION].```

I use rvm and do not have that environment variable set.  When I switch the conditional in the Gemfile to be block based instead of inline, the warning / message seems to go away.

Thoughts?

**Comment by stevecrozz** [comment]
Looks like RVM attempts to figure out which version of ruby you want to use by parsing it out of the Gemfile with some dirty black shell magic: https://github.com/wayneeseguin/rvm/blob/f01bfa7959cca7188caf59b094cfee33e0e1503c/scripts/functions/rvmrc_project#L133-L137

You can see exactly what's happening when you run it by hand:
```bash
$ echo "ruby ENV['GEMFILE_RUBY_VERSION'] if ENV['GEMFILE_RUBY_VERSION']" | command tr -d '\r' | sed -n "s/[[:space:]]+rescue[[:space:]]+nil$//; /^\s*ruby[[:space:](]/ {s/^\s*ruby//; s/[[:space:]()'\"]//g; p;}"
ENV[GEMFILE_RUBY_VERSION]ifENV[GEMFILE_RUBY_VERSION]
```
Splitting it over multiple lines isn't a great fix because the result is still wrong, even if RVM doesn't choke:
```bash
$ echo $'if ENV["GEMFILE_RUBY_VERSION"]\n  ruby ENV["GEMFILE_RUBY_VERSION"]\nend' | command tr -d '\r' | sed -n "s/[[:space:]]+rescue[[:space:]]+nil$//; /^\s*ruby[[:space:](]/ {s/^\s*ruby//; s/[[:space:]()'\"]//g; p;}"
ENV[GEMFILE_RUBY_VERSION]
```

This seems to me like a bug in RVM. We can potentially work around it with something like:
```ruby
send(:ruby, ENV['GEMFILE_RUBY_VERSION']) if ENV['GEMFILE_RUBY_VERSION']
```

You're welcome to give that a try and we can make that change, but it's not awesome that we're essentially relying on this shell scripting snippet in order to support RVM. Maybe you can report this as a bug with them...

stevecrozz:
[...]

PR #1528, opened 2022-07-07
Remove GEMFILE_RUBY_VERSION

biow0lf:
Remove `GEMFILE_RUBY_VERSION` env support.

End users should not set ruby version.


biow0lf:
@stevecrozz just one small step to the future

Gemfile
```diff
@@ -3,7 +3,7 @@ source 'https://rubygems.org'
 RAILS_VERSION = '~> 4.2.11'
 # RAILS_VERSION = '5.0.7.2'
 
-send :ruby, ENV['GEMFILE_RUBY_VERSION'] if ENV['GEMFILE_RUBY_VERSION']
+ruby File.read(".ruby-version")
```
stevecrozz, line 6: That old insanity was here due to an ancient issue with rvm:
https://github.com/errbit/errbit/issues/838

Do you know if it's been fixed sometime in the past 7 years?

biow0lf: I will check and update PR if needed.

I look on rvm source code and I don't understand how it working under the hood. So, I will install VM and rvm inside to test.

2called-chaos:
Just my 2 cents but .ruby-version is not in the repo and is gitignored. Should we add this then and or make this failsafe?

biow0lf:
> Just my 2 cents but .ruby-version is not in the repo and is gitignored. Should we add this then and or make this failsafe?

@2called-chaos

this PR add `.ruby-version` file to repo and git-un-ignore it.

Gemfile
biow0lf: Ok. I am back. This year is crazy hard. It starts for me and my family in Mariupol, Ukraine. [The Washington Post on Mariupol news](https://www.washingtonpost.com/search/?query=mariupol&btn-search=). Running from war and writing open source it hard. So, I am trying. :)

So. I check rvm. And, let me describe what I found and think about all this.

Let's talk about `.ruby-version` file syntax and history. Everything start with [`rbenv`](rbenv/rbenv). `rbenv` is simple wrapper for bash to run ruby (and compile via `ruby-build`). Syntax is straightforward. e.g. `3.1.2`. Which tell rbenv wrapper to switch ruby version to `3.1.2`. Nothing more, nothing less.

Otherwise, rvm use (maybe, *uses in past*) `.rvmrc` which is bash script (this is way it hang terminal sometimes). And here starts all this voodoo-magic with parsing something in Gemfile, etc. rvm already understand `.ruby-version` syntax and just works. Even more, if rvm see `.ruby-version` file it will not try find ruby version anywhere more. And use ruby version from `.ruby-version` file. Just like `rbenv` do.

All ruby version managers now understand `.ruby-version` syntax and just switch ruby. This code is legacy and should be removed.

One more thing: I still not sure about ruby version (e.g. `ruby 3.1.2`) in Gemfile itself. Right now, I think, it should be. Looks like bundler ruby gems version solver use this information.

So, this is just legacy code that should be updated.

stevecrozz:
Thanks for the extra diligence!

stevecrozz (approved):

PR #1530, opened 2022-07-30
Upgrade rails to 5.0

biow0lf:
Aloha!

Main change: Upgrade Ruby on Rails to 5.0

Changes:
1. Update rubygems to 3.3.19
2. Update bundler to 2.3.19
3. Add `activemodel-serializers-xml` gem. This functionality was removed from rails 5.0. We use it `.to_xml`.
4. `mongoid` was updated to `6.0.3`. This is latest version that works with rails 5.0.
5. `yajl-ruby` gem was removed without replacement. If we need more speed, we can look at `oj` gem and `ox` gem for xml parsing.
6. Add `rails-controller-testing` gem. Required by rspec.
7. `rails_12factor` gem removed without replacement. All functionality was merged in rails itself.
8. Some gems were updated. Commit by commit. Before rails upgrade.
9. `.with(consistency: :strong)` was removed. `mongoid` gem deprecate them and remove support. We can use replacement to read from main mongo db. Did we really need this?
10. Replace `render text:` with `render body:`.
11. Remove `ActionController::RedirectBackError`. This exception deprecated.
12. Replace `update_attributes` with just `update`. `update_attributes` is just long alias to `update`.
13. `redirect_to :back` was replaced with `redirect_back fallback_location: root_path`
14. `Watcher#user` should be explicitly marked as `optional`.
15. `autotest/discover.rb` was removed without replacement. I think, this is ancient relict.
16. Rails files was synced with rails 5.0.
17. Add empty `public/apple-touch-icon-precomposed.png` to reduce error logs.
18. Add empty `public/apple-touch-icon.png` to reduce error logs.
19. Fix rails deprecations with positional arguments in tests. Like: before `get :show, id: app.id` and after `get :show, params: { id: app.id }`.
20. Rails 5.0 more strict about status code. You can't use `:error`. You should exactly code like `:internal_server_error`.
21. Rails rename `serve_static_files` to `public_file_server.enabled`. I left env name `Errbit::Config.serve_static_assets`.
22. Rails configs were synced with rails defaults. With comments and other stuff inside.
23. Sync `public/404.html`, `public/422.html` and `public/500.html` with rails 5.0.
24. Remove monkey-patch from `spec/spec_helper.rb`
25. Fix deprecation in `add_on_blank(:resolved)`
26. Fix `spec/requests/health_controller_spec.rb`

Ok, it's time to talk about broken mongoid. [Link to issue](https://jira.mongodb.org/browse/MONGOID-4499). Test suite pass, so.

I left:
```
# TODO: uncomment with new mongoid 6.2.x or later
# app.watchers.build(email: 'foo@example.com')
# app.save
```

with rails 5.1 should work properly.

P.S.: Running test suite generates a lot of warnings. This is not good, but... We can fix them later. :)

P.S.2: I update Dockerfile. And build image [here](https://hub.docker.com/r/biow0lf/errbit).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -87,7 +87,6 @@ group :test do
 end
 
 group :heroku, :production do
-  gem 'rails_12factor', require: ENV.key?("HEROKU")
 end
 
 group :no_docker, :test, :development do
--- a/Gemfile
+++ b/Gemfile
@@ -49,7 +49,6 @@ gem 'httparty'
 gem 'flowdock'
 
 gem 'ri_cal'
-gem 'yajl-ruby', platform: 'ruby'
 gem 'json', platform: 'jruby'
 
 # For Ruby 2.7
--- a/Gemfile
+++ b/Gemfile
@@ -64,6 +64,7 @@ group :development, :test do
 end
 
 group :development do
+  gem 'listen', '~> 3.0.5'
   gem 'better_errors'
   gem 'binding_of_caller', platform: 'ruby'
   gem 'meta_request'
--- a/Gemfile
+++ b/Gemfile
@@ -87,9 +87,6 @@ group :test do
   gem 'coveralls', require: false
 end
 
-group :heroku, :production do
-end
-
 group :no_docker, :test, :development do
   gem 'mini_racer', platform: :ruby # C Ruby (MRI) or Rubinius, but NOT Windows
 end
--- a/Gemfile
+++ b/Gemfile
@@ -8,6 +8,7 @@ gem 'actionmailer', RAILS_VERSION
 gem 'actionpack', RAILS_VERSION
 gem 'railties', RAILS_VERSION
 
+gem 'activemodel-serializers-xml'
 gem 'actionmailer_inline_css'
 gem 'decent_exposure'
 gem 'devise'
--- a/Gemfile
+++ b/Gemfile
@@ -70,6 +70,7 @@ group :development do
 end
 
 group :test do
+  gem 'rails-controller-testing'
   gem 'rake'
   gem 'rspec'
   gem 'rspec-rails', require: false
--- a/app/controllers/site_config_controller.rb
+++ b/app/controllers/site_config_controller.rb
@@ -6,8 +6,8 @@ class SiteConfigController < ApplicationController
   end
 
   def update
-    SiteConfig.document
-      .update(notice_fingerprinter: filtered_update_params.to_h)
+    SiteConfig.document.
+      update(notice_fingerprinter: filtered_update_params.to_h)
 
     flash[:success] = 'Updated site config'
 
@@ -17,9 +17,9 @@ class SiteConfigController < ApplicationController
 private
 
   def filtered_update_params
-    params.require(:site_config)
-      .require(:notice_fingerprinter_attributes)
-      .permit(:error_class, :message, :backtrace_lines, :component, :action,
-        :environment_name)
+    params.require(:site_config).
+      require(:notice_fingerprinter_attributes).
+      permit(:error_class, :message, :backtrace_lines, :component, :action,
+             :environment_name)
   end
 end
--- a/app/controllers/site_config_controller.rb
+++ b/app/controllers/site_config_controller.rb
@@ -17,9 +17,15 @@ class SiteConfigController < ApplicationController
 private
 
   def filtered_update_params
-    params.require(:site_config).
+    params.
+      require(:site_config).
       require(:notice_fingerprinter_attributes).
-      permit(:error_class, :message, :backtrace_lines, :component, :action,
-             :environment_name)
+      permit(
+        :error_class,
+        :message,
+        :backtrace_lines,
+        :component,
+        :action,
+        :environment_name)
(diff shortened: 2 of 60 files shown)


biow0lf:
@stevecrozz it is time to ask about code review.

One test if failing. Plus, a lot of deprecations. But, anyway, it's time to show and ask feedback.

Gemfile
```diff
@@ -21,8 +21,7 @@ gem 'haml'
 gem 'htmlentities'
 gem 'kaminari'
 gem 'kaminari-mongoid'
-gem 'mongoid', '~> 5.4'
-# gem 'mongoid', '6.0.3'
+gem 'mongoid', '6.0.3'
```
stevecrozz, line 24: Can we make this more flexible? like ~> 6.0?

app/controllers/site_config_controller.rb
```diff
@@ -6,24 +6,20 @@ def index
...
+
     redirect_to action: :index
   end
 
 private
 
   def filtered_update_params
-    params.
-      require(:site_config).
+    params.require(:site_config).
```
stevecrozz, line 20: Why is this syntax changing?

config/environments/development.rb
```diff
@@ -9,13 +9,28 @@
...
   config.eager_load = false
 
-  # Show full error reports and disable caching.
-  config.consider_all_requests_local       = true
-  config.action_controller.perform_caching = false
+  # Show full error reports.
+  config.consider_all_requests_local = true
+
+  # Enable/disable caching. By default caching is disabled.
+  if Rails.root.join('tmp', 'caching-dev.txt').exist?
```
stevecrozz, line 16: Why do we need this change?

config/environments/development.rb
```diff
@@ -27,11 +42,13 @@
...
-  config.assets.raise_runtime_errors = false
+  # Suppress logger output for asset requests.
+  config.assets.quiet = true
 
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
+
+  # Use an evented file watcher to asynchronously detect changes in source code,
+  # routes, locales, etc. This feature depends on the listen gem.
+  # config.file_watcher = ActiveSupport::EventedFileUpdateChecker
```
stevecrozz, line 53: Seems strange to have this added, but commented out.

config/environments/production.rb
```diff
@@ -56,12 +64,19 @@
...
 
-  # Disable automatic flushing of the log to improve performance.
-  # config.autoflush_log = false
-
   # Use default logging formatter so that PID and timestamp are not suppressed.
   config.log_formatter = ::Logger::Formatter.new
 
+  # Use a different logger for distributed setups.
+  # require 'syslog/logger'
+  # config.logger = ActiveSupport::TaggedLogging.new(Syslog::Logger.new 'app-name')
```
stevecrozz, line 72: Can you remove all the lines that were added but commented out? I don't know what they are and if we leave them commented, then we clearly don't need them :)

config/environments/development.rb
burnettk: It’s the built-in `app:update` command, I think. I agree that it’s suboptimal for that thing to add commented out lines. I guess the rails folks wanted to make users aware of the options and expected them to manually make updates. I’m lamenting not having open sourced our rails upgrade scripts before we stopped focusing on rails stuff, since they were significantly better than the framework-provided commands.

app/controllers/site_config_controller.rb
biow0lf: I debug it. Trying to understand why it not work and reformat. I will rollback.

config/environments/development.rb
biow0lf: @stevecrozz to be sync with Rails defaults.

config/environments/development.rb
biow0lf: @stevecrozz strange? Ok, it will be removed in next rails upgrade. But, If you like, I will add listen gem and everything around.

@burnettk Yep, you are completely right. This is `app:update` rake task.

config/environments/production.rb
biow0lf: @stevecrozz this is default rails settings. I prefer to leave them as-is. Just to make diff less while upgrading.

config/environments/development.rb
biow0lf: @stevecrozz I add listen and uncomment this code.

Gemfile
biow0lf: @stevecrozz this gem update fills fragile. So, it locked for future. Anyway, 6.0.3 is last version that works with rails 5.0.

I hope, I answer all questions?

biow0lf:
Next update will be with fixes for deprecations and upgrade `omniauth` and related gems.

stevecrozz:
Looks pretty good to me @biow0lf . Are you ready?

biow0lf:
@stevecrozz I am ready. Let's merge and see.

stevecrozz (approved):

stevecrozz:
🎉

PR #1531, opened 2022-08-22
Preparing Rails upgrade

biow0lf:
List of changes:

1. Fix rails deprecations
2. Add GitHub Action for run rspec
3. Add GitHub Action for run rubocop
4. Add mongodb 6.0 to list of versions to test with (to new GitHub Action)
5. Update rubygems/bundler versions
6. All GitHub Actions run by cron for regular checking
7. Bump copyright year 🔥 
8. Major `omniauth` and related gems upgrade.
9. Add `omniauth-rails_csrf_protection`. All sign in with "provider" are via POST with CSRF token checks.
10. Minor gems updates. Nothing interesting to describe.
11. Rubocop TODO config regenerated. Mostly, I only change `Metrics/AbcSize:` from `52` to `54`. Everything else rubocop make itself. I don't want to change any rubocop changes. Any changes to rubocop rules always controversial and require to much discussion.

I prefer to think that from now, Circle CI config deprecated. And will be removed in future. GitHub Actions already mature solution. And Circle CI don't build their new images for MongoDB. I don't know how to add this new release to check. And a lot of projects just migrate from Circle CI to GitHub Actions.


biow0lf:
@stevecrozz this PR is ready for review 🎆

biow0lf:
@stevecrozz

.github/workflows/rspec.yml
```diff
@@ -0,0 +1,32 @@
...
+  schedule:
+    - cron: "0 21 * * 6"
+
+jobs:
+  rspec:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        mongodb-version: ["4.0", "4.2", "4.4", "5.0", "6.0"]
```
stevecrozz, line 19: Is this list of mongo versions one that you pulled from somewhere else? I think its probably overkill here and a bit wasteful. If the idea of this is to be a canary for upstream changes that break errbit, maybe we want to run this job on only the 'latest' mongo somehow..

.github/workflows/rubocop.yml
```diff
@@ -0,0 +1,23 @@
...
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+  schedule:
+    - cron: "0 21 * * 6"
```
stevecrozz, line 11: This one I don't think benefits from a scheduled run does it? Is there a good reason to run rubocop on a schedule?

.github/workflows/rspec.yml
biow0lf: I take it from Circle CI config. And add 6.0. Circle CI uses different mongodb images. Their own. Right now, they are deprecated. I don’t want to change or update circle ci config. We should remove it. But for now, it has some external test coverage. I will take care about this later.

We can remove older versions of mongodb. But, lets discuss this in another PR. Plus, I run errbit with mongodb 4.0.

.github/workflows/rubocop.yml
biow0lf: Yes. Check that everything for rubocop is installing and running. Once per week is good trade off.

biow0lf: Even better. We will know that this GiHub Action still working.

stevecrozz: Ok. We'll see how it goes

stevecrozz (approved):

biow0lf:
@stevecrozz looks like something go wrong with Circle CI. Can you re-trigger run?

stevecrozz:
Looks like I cannot. Not sure what's going on there.
Screen Shot 2022-09-12 at 10 14 34 AM

PR #1539, opened 2023-03-23
Merge errbit upstream

willtcarey:
Merges errbit upstream to fix the mime_magic error and we'll see if it fixes the SSL issue


willtcarey:
Sorry, wrong repo!

PR #1544, opened 2023-08-16
Handle strings in pretty_hash to prevent exceptions (NoMethodError: undefined method 'keys' for #<String>)

jpeterburs:
We are experiencing errors in `Self.Errbit` when trying to open some Problems.
After some digging, I found out that `notice.params` returned a String, instead of a hash.

Problem
Backtrace

### DIFF
--- a/app/helpers/hash_helper.rb
+++ b/app/helpers/hash_helper.rb
@@ -1,6 +1,6 @@
 module HashHelper
   def pretty_hash(hash, nesting = 0)
-    return '{}' if hash.empty?
+    return '{}' if hash.empty? || hash.is_a?(String)
 
     tab_size = 2
     nesting += 1
--- a/spec/helpers/hash_helper_spec.rb
+++ b/spec/helpers/hash_helper_spec.rb
@@ -1,0 +1,30 @@
+describe HashHelper do
+  describe '.pretty_hash' do
+    let(:expected_pretty_hash) do
+      <<~HASH.chomp
+        {
+          "action" => "some_action",
+          "controller" => "some_controller"
+        }
+      HASH
+    end
+
+    it 'is expected to prettify hashes' do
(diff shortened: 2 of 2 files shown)


stevecrozz:
I don't mind the idea, but according to Airbrake API v3, the params should be an object and not a string when reporting an error. Wonder what happens if you tried to send `params: ''` to Airbrake...

jpeterburs:
> Wonder what happens if you tried to send `params: ''` to Airbrake...

In a rails console, I've tried `Airbrake.notify('This is a test', params: '')` and got the following result:
image

Digging around in https://github.com/airbrake/airbrake-ruby I got the following backtrace:
- [Airbrake.notify](https://github.com/airbrake/airbrake-ruby/blob/f2cc9917a18004330d17a05c7ab4f84c1ea2ce02/lib/airbrake-ruby.rb#L182)
- [Airbrake::NoticeNotifier#notify](https://github.com/airbrake/airbrake-ruby/blob/f2cc9917a18004330d17a05c7ab4f84c1ea2ce02/lib/airbrake-ruby/notice_notifier.rb#L34)
- [Airbrake::NoticeNotifier#send_notice](https://github.com/airbrake/airbrake-ruby/blob/f2cc9917a18004330d17a05c7ab4f84c1ea2ce02/lib/airbrake-ruby/notice_notifier.rb#L107)
- [Airbrake::NoticeNotifier#build_notice](https://github.com/airbrake/airbrake-ruby/blob/f2cc9917a18004330d17a05c7ab4f84c1ea2ce02/lib/airbrake-ruby/notice_notifier.rb#L54)

In this backtrace, I couldn't find any handling of strings, but my sting was handled anyhow...

I've looked at the notice that caused this again, and noticed that the notice in question was likely triggered by JavaScript.
Sadly I'm unable to debug our reporter for JavaScript based exceptions, but I think, that it is able to send strings.

stevecrozz:
Looks good to me. We can accept strings here even though the client is probably behaving badly.

PR #1546, opened 2023-08-26
Remove need for `include Mongoid::Attributes::Dynamic`

sahglie:
`doc.unset(field)` in mongoid >= 7.x interprets this as setting an attribute and since there is no attribute called "#{key}.#{digest}" it fails. Previous version of mongoid appear to just update the hash that #{key} pointed to. So, looks like we just need to remove the cached digest from the hash using a different method. This would do it:

```
if (doc[k].try(:[], digest).try(:[], :count)).to_i > 1
  doc.inc("#{field}.count" => -1)
else
  h = doc[k] || {}
  h.delete(digest)
  doc.set("#{k}": h)
end
```

With this update and removing `include Mongoid::Attributes::Dynamic` from `Problem` class the entire test suite passes.


biow0lf:
I believe that everything from this PR already in merged in main. So, I am close it now.

PR #1547, opened 2023-09-03
Feature that temporarily mutes problems silencing error notifications

sahglie:


burnettk (approved):
did not run the app, but code looks good

PR #1573, opened 2025-02-28
Remove test-queue and update docs

biow0lf:
On my machine (MacBook Pro on M1 Pro), tests already running 2 minutes.


burnettk:
seems like a reasonable tradeoff, thanks.

PR #1575, opened 2025-03-03
Remove USER_GEMFILE env support

biow0lf:
This PR removes support of custom `UserGemfile` support.

From now, only `UserGemfile` is supported.

Why? Dependabot can't parse `Gemfile` with custom ruby code. So, I simplify this code.  Users still can use `UserGemfile` and Dependabot can parse our `Gemfile` and update dependencies.


biow0lf:
We need to upgrade a lot of ruby dependencies. But, manually upgrade 30-40+ gems is hard. Let's Dependabot make this dirty job.

PR #1896, opened 2025-04-18
Upgrade airbrake gem

biow0lf:


mikelolasagasti:
Done in https://github.com/errbit/errbit/pull/3076

PR #2634, opened 2026-02-06
OIDC aka OpenID Connect

biow0lf:
WIP

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -302,6 +302,24 @@ Devise.setup do |config|
       google_options
   end
 
+  if Errbit::Config.oidc_authentication || Rails.env.test?
+    config.omniauth :openid_connect, {
+      name: ENV.fetch("OIDC_NAME", nil).to_sym,
+      issuer: ENV.fetch("OIDC_ISSUER", nil),
+      scope: ENV.fetch("OIDC_SCOPES", "openid,profile,email").split(",").map(&:to_sym),
+      discovery: true,
+      response_type: :code,
+      client_options: {
+        port: 443,
+        scheme: "https",
+        host: ENV.fetch("OIDC_HOST", nil),
+        identifier: ENV.fetch("OIDC_CLIENT_ID", nil),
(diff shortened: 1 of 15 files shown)


docs/oidc.md
```diff
@@ -0,0 +1,34 @@
...
+Default: `[]`
+
+## Example configurations
+
+### GitLab.com or any gitlab-powered instance
+
+```shell
+export OIDC_ENABLED="true"
+export OIDC_NAME="gitlab"
+export OIDC_SCOPES="openid,email,read_user"
```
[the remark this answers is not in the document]
biow0lf: > `read_user` is not a valid scope; use `profile`.

This is example configuration that works with GitLab.

dchandekstark:
Auto-provisioning is convenient, and I use it for other apps, but it might be good in general to toggle that feature.

config/initializers/devise.rb
```diff
@@ -302,6 +302,24 @@
       google_options
   end
 
+  if Errbit::Config.oidc_enabled || Rails.env.test?
+    config.omniauth :openid_connect, {
+      name: ENV.fetch("OIDC_NAME", nil).to_sym,
```
dchandekstark: `name` isn't required and will caused a nil error here if not set.

config/initializers/devise.rb
```diff
@@ -302,6 +302,24 @@
...
+      issuer: ENV.fetch("OIDC_ISSUER", nil),
+      scope: ENV.fetch("OIDC_SCOPES", "openid,profile,email").split(",").map(&:to_sym),
+      discovery: true,
+      response_type: :code,
+      client_options: {
+        port: 443,
+        scheme: "https",
+        host: ENV.fetch("OIDC_HOST", nil),
+        identifier: ENV.fetch("OIDC_CLIENT_ID", nil),
+        secret: ENV.fetch("OIDC_SECRET", nil),
```
dchandekstark: Ditto as above comment re: identifier.

PR #2723, opened 2026-03-21
refactor: remove decent_exposure gem in favor of plain Rails helpers

### LINKED ISSUE #1855: Remove `decent_exposure` gem
**Why?**

It doesn't work on our case.

**Test plan:**

1. Rewrite code without `decent_exposure`
2. Remove `decent_exposure` gem
3. Check that everything still works

mvanhorn:
## Summary

Removes the `decent_exposure` gem and replaces all `expose` calls with standard Rails private methods + `helper_method`.

## Why this matters

Per #1855, `decent_exposure` doesn't work for errbit's use case. Replacing it with plain Rails patterns removes a dependency and makes the controller logic explicit.

## Changes

- `Gemfile`: removed `gem "decent_exposure"`
- `app/controllers/apps_controller.rb`: 7 `expose` calls replaced with private memoized methods
- `app/controllers/comments_controller.rb`: 3 `expose` calls replaced
- `app/controllers/problems_controller.rb`: 7 `expose` calls replaced
- `app/controllers/concerns/problems_searcher.rb`: 5 `expose` calls replaced, using `included do` block for `helper_method`

All method names are unchanged so views require no modifications. Each method uses `@var ||=` memoization to match `expose`'s lazy evaluation behavior.

## Testing

Verified no remaining `decent_exposure` references in Ruby files via `grep -rn "decent_exposure" --include="*.rb"`.

Fixes #1855

This contribution was developed with AI assistance (Claude Code).


app/controllers/apps_controller.rb
```diff
@@ -113,6 +81,42 @@ def search
...
 
+  def app_scope
+    @app_scope ||= params[:search].present? ? App.search(params[:search]) : App.all
+  end
+
+  def apps
+    @apps ||= app_scope.to_a.sort.map { |app| AppDecorator.new(app) }
+  end
+
+  def app
```
mikelolasagasti, line 92: `new` and `create` have no app ID, so this calls `App.find(nil)` and raises instead of building a new app. Please return a new `App` for the create path.

app/controllers/comments_controller.rb
```diff
@@ -28,6 +26,18 @@ def destroy
...
+  def app
+    @app ||= App.find(params[:app_id])
+  end
+
+  def problem
+    @problem ||= app.problems.find(params[:problem_id])
+  end
+
+  def comment
+    @comment ||= Comment.new(comment_params)
```
mikelolasagasti, line 38: `destroy` needs the existing comment, but this always builds a new one. The requested comment is never loaded or deleted, and `comment_params` can raise because destroy requests do not include `params[:comment]`.

app/controllers/apps_controller.rb
```diff
@@ -7,39 +7,7 @@ class AppsController < ApplicationController
...
-      pr.in_env(
-        params[:environment]
-      ).ordered_by(params_sort, params_order).page(params[:page]).per(current_user.per_page)
-    end
-  end
-
-  expose(:users) do
-    User.all.sort_by { |u| u.name.downcase }
-  end
+  helper_method :app_scope, :apps, :app, :app_decorate, :all_errs, :problems, :users
```
mikelolasagasti, line 10: These methods are now defined under `private`, but existing specs call methods such as `controller.app` and `controller.problems` directly. Those specs will fail with private-method errors. Please keep these methods public or update the specs.

mikelolasagasti (changes requested):
The PR currently has two action-level regressions: app creation/new actions resolve an existing record, and comment deletion builds a new record instead of loading the target. It also makes exposed methods private while existing specs call them directly. Please address these before merging.

mvanhorn:
Good catches, both were real. Fixed in this push:

`apps_controller.rb:92` - `app` no longer calls `App.find(nil)` on the create path. It returns `App.find(...)` when an id is present and otherwise builds a new `App`, empty for `new` and from `app_params` for `create`. Added a spec that posts to `create` without stubs and asserts the record is built and persisted; it fails on the old code.

`comments_controller.rb:38` - `comment` now loads `Comment.find(params[:id])` when an id is present and only builds from `comment_params` otherwise, so `destroy` operates on the requested comment and no longer raises on the missing `params[:comment]`. The existing destroy specs were failing on this exact error and now pass.

`apps_controller.rb:10` - moved `private` below the exposed methods in `apps_controller`, `comments_controller` and `problems_controller` so `controller.app`, `controller.problems` and friends stay callable from the specs. In `ProblemsSearcher` I dropped `private` entirely: all five methods there are declared `helper_method`, so private was wrong for them regardless.

Also pruned the leftover `decent_exposure` entries from `Gemfile.lock`, which the Gemfile removal had left behind.

`bundle exec rspec spec/controllers/` is green: 176 examples, 0 failures. I checked both regressions are load-bearing by reverting each controller in turn; without the fix the comment specs fail with the `params.require(:comment)` error you predicted and the app specs fail in `App.find`.

mvanhorn:
Rebased on main as well, since the branch had drifted since March. Two things worth flagging from the conflict resolution:

`problems_controller.rb` - main had moved these lookups to Rails 8 `params.expect(...)` inside the `expose` blocks. The extracted plain methods now use `params.expect` too, so `app`, `app_scope` and `problem` keep the same strong-params behavior they had under decent_exposure rather than silently reverting to `params[...]`.

`Gemfile.lock` - took main's `devise 5.0.4` and the 4.0.6 ruby pin, and dropped only the `decent_exposure` entries.

`bundle exec rspec spec/controllers/` is still 176 examples, 0 failures on the new base.

mikelolasagasti (approved):
LGTM, thanks for addressing the reported issues.

biow0lf:
> Removes the `decent_exposure` gem and replaces all `expose` calls with standard Rails private methods + `helper_method`.

"Standard" "Rails" is just action in controller that loads data. e.g.:

```ruby
class AppsController
  def index
    @apps = App.page(params[:page])
  end
end
```

mvanhorn:
Fair point, @biow0lf. I went with memoized methods plus `helper_method` so the views could keep calling `app`, `apps`, `problems` and the rest without touching every template, but you are right that it keeps decent_exposure's lazy-lookup shape rather than plain Rails. If you would rather have each action assign its own instance variables (`@apps = App.page(...)` in `index`, `@app = App.find(...)` in the member actions) with the views reading `@app` / `@apps`, I am happy to rework this PR that way. It reaches into the views and `problems_searcher` too, so I wanted to confirm that is the direction you want before doing it here.

biow0lf:
> so I wanted to confirm that is the direction you want before doing it here.

Yes

mvanhorn:
Reworked as discussed, thanks for confirming the direction.

Each action now loads what it needs into an instance variable and the views read those directly. `index` assigns `@apps`, the member actions assign `@app`, and the same for `@problems`, `@all_errs`, `@users` and the rest. The `helper_method` declarations and the memoized private readers are gone, and `problems_searcher` works against the instance variables now.

33 files: 4 controllers/concerns, 14 views and builders, and the specs that referenced the old readers.

Full suite is green locally, 702 examples / 0 failures, run against MongoDB rather than mocked out.

One thing worth flagging: moving the values out of `helper_method` means `apps_helper` and `sort_helper` now read `@params_order` and `@all_errs` directly, which trips `Rails/HelperInstanceVariable` (11 new offenses in `app/`). Your rubocop workflow only lints an explicit file list that none of these are on, so CI is unaffected, but if you would rather those helpers took the values as arguments instead, say the word and I will pass them in.

PR #3077, opened 2026-08-13
Fix headless Chrome system specs in containers

mikelolasagasti:
Add the Chrome flags `--no-sandbox` and `--disable-dev-shm-usage`, which are commonly required when running headless Chrome in Linux containers.

This allows the existing Selenium/Capybara configuration to work in containerized environments without changing the selected browser or driver.

### DIFF
--- a/spec/support/capybara.rb
+++ b/spec/support/capybara.rb
@@ -8,6 +8,10 @@ Capybara.register_driver :selenium_chrome_headless do |app|
 
   browser_options = Selenium::WebDriver::Chrome::Options.new.tap do |opts|
     opts.add_argument("--headless=new")
+    # Chrome's sandbox often cannot initialize inside Docker/Podman test containers.
+    opts.add_argument("--no-sandbox")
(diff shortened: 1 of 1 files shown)


spec/support/capybara.rb
```diff
@@ -8,6 +8,8 @@
 
   browser_options = Selenium::WebDriver::Chrome::Options.new.tap do |opts|
     opts.add_argument("--headless=new")
+    opts.add_argument("--no-sandbox")
+    opts.add_argument("--disable-dev-shm-usage")
```
biow0lf, line 14: Can you add comments why we need those options?

Anyway, looks good! Thanks!

biow0lf (changes requested):
Add comments why we need those options for google chrome

mikelolasagasti:
Added a couple of short comments in the code. Is it OK this way?

biow0lf (approved):

biow0lf:
> Added a couple of short comments in the code. Is it OK this way?

Yep.

PR #3076, opened 2026-08-12
Bump airbrake from 4.3.8 to 13.0.5

mikelolasagasti:
Airbrake 13 no longer exposes the old XML notice APIs used by the `ErrorReport` spec, including `Notice#to_xml` and the associated internal classes.

Update the spec to generate a notice with the current Airbrake gem and parse it through Errbit's Airbrake v3 parser. This preserves the compatibility boundary between Airbrake-generated payloads and Errbit while removing the test-only dependency on Airbrake 4 internals.

Legacy XML fixture coverage remains unchanged.

Fixes #1849

PR #3078, opened 2026-08-13
Replace hoptoad_notifier self-reporting with airbrake

### LINKED ISSUE #1850: Replace `hoptoad_notifier` with `airbrake` gem
**What is `hoptoad_notifier`?**

`hoptoad_notifier` is old name for `airbrake` gem.

**Why we use `hoptoad_notifier`?**

We use `hoptoad_notifier` inside Errbit for self-tracking errors from Errbit itself.

**Why we want to replace it?**

`hoptoad_notifier` is old gem. It's only a matter of time, when it breaks.

**Test plan:**

1. Replace `hoptoad_notifier` with `airbrake`.
2. Check that Errbit still can track self errors in itself

mikelolasagasti:
(built on top of https://github.com/errbit/errbit/pull/3076)

Remove the `hoptoad_notifier` dependency and replace its monkeypatched self-error reporting path with `errbit::SelfErrorReporter`.
    
The new reporter builds Airbrake v3 notices locally with `Airbrake.build_notice`, injects the `Self.Errbit` app API key, and persists them through the existing v3 parser and `ErrorReport` pipeline. It preserves the previous non-public environment behavior, continues to ignore `Mongoid::Errors::DocumentNotFound`, and never invokes Airbrake's delivery APIs.
    
Existing `Self.Errbit` application settings are now preserved instead of resetting the GitHub repository on every internal error.
    
Fixes #1850


lib/errbit/self_error_reporter.rb
```diff
@@ -0,0 +1,131 @@
+# frozen_string_literal: true
+
+require "airbrake"
+require "airbrake/rack/middleware"
+
+module Errbit
+  module SelfErrorReporter
+    SELF_APP_NAME = "Self.Errbit"
+    # Preserve the previous notifier's public-environment behavior.
+    NON_PUBLIC_ENVIRONMENTS = ["development", "test", "cucumber"].freeze
```
biow0lf: Remove "cucumber".

spec/lib/errbit/self_error_reporter_spec.rb
```diff
@@ -0,0 +1,147 @@
...
+    end
+  end
+
+  after { Airbrake.reset }
+
+  before do
+    allow(described_class).to receive(:public_environment?).and_return(true)
+  end
+
+  %w[development test cucumber].each do |environment|
```
biow0lf: Style: I use only `["str"]` syntax.

Remove "cucumber".

biow0lf (approved):
Few notes. Merge it.

PR #3091, opened 2026-08-19
Enable CSP

biow0lf:


mikelolasagasti:
Implemented in https://github.com/errbit/errbit/pull/3109

PR #3108, opened 2026-08-26
Configure HTTP Feature Policy

biow0lf:


mikelolasagasti:
Agree with Copiltot. This feature used to be named `Feature Policy`, but even W3 mentions that it was renamed to `Permissions Policy`

https://www.w3.org/TR/permissions-policy/#introduction

Otherwise LGTM

PR #3123, opened 2026-09-01
Drop the stray quote after the request url in issue reports

bmwiedemann:
With spec coverage to prevent regression.

### DIFF
--- a/app/views/issue_trackers/markdown.erb
+++ b/app/views/issue_trackers/markdown.erb
@@ -4,7 +4,7 @@
 ## Summary ##
 <% if notice.request['url'].present? %>
 ### URL ###
-[<%= notice.request['url'] %>](<%= notice.request['url'] %>)"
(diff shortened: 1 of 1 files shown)


biow0lf:
@bmwiedemann 

Nice catch!

spec/views/issue_trackers/markdown.erb_spec.rb
biow0lf, line 1: I don't think that we are need tests for quote. Drop them and I will merge.

biow0lf (changes requested):
I don't think that we are need tests for quote. Drop them and I will merge.

spec/views/issue_trackers/markdown.erb_spec.rb
bmwiedemann: Done.

biow0lf (approved):

PR #3124, opened 2026-09-01
Accept public_repo scope in User#can_create_github_issues?

bmwiedemann:
as documented in the README.

Also add spec coverage for it.

### DIFF
--- a/app/controllers/problems_controller.rb
+++ b/app/controllers/problems_controller.rb
@@ -7,6 +7,8 @@ class ProblemsController < ApplicationController
     :resolve_several, :unresolve_several, :unmerge_several
   ]
 
+  before_action :need_github_issue_permission, only: [:create_issue, :close_issue]
+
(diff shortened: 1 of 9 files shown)


mikelolasagasti:
This change looks correct by itself and is consistent with what the README says. However, `can_create_github_issues?` is not used anywhere in the application. Therefore, accepting public_repo does not change how Errbit works today.

Is using this check when creating issues intentionally left for a future pull request? If so, please mention that in the PR description. Otherwise, please add that integration to this pull request.

bmwiedemann:
It is part of a larger WIP rework to allow to file Github issues without giving full write permissions to the app.
I split two independent fixes from the larger work that put it to use.

Much of the coding was done by ClaudeAI with an extra review round that found these to be correct. The changes look OK to me, though my own RoR knowledge is rusty and I did not test the split changes in a deployment. So give it some extra care (if you got the time).

PR #3127, opened 2026-09-02
Persist active tab in URL hash across notice pagination

espen:
Often I want to compare URL params for a group of logged errors. This was tedious as I had to click previous error, switch tab, click previous error, switch tab. This PR will keep focus on selected tab. I am not familiar with pjax so had Claude arrange this PR for me. Let me know if there is a preferred different approach.

By Claude:
The tab links on the problem page set a fragment (#params, #backtrace, etc.) that was never read back, so opening a link with a hash or paging through notices with Older/Newer always reset the view to Summary and dropped the hash from the URL.

Initialize the current tab from location.hash on load, write the hash via history.replaceState when a tab is activated, and re-append it after pjax pagination so the URL stays shareable. Unknown hashes fall back to the Summary tab.

Also fix the pjax setup for the vendored jquery-pjax API: the container must be passed as a string selector ($(document).pjax(links, '#content')) instead of calling .pjax() on the container element, which threw "expected string value for 'container' option" and fell back to full page loads on every pagination click. The replaceState calls preserve history.state so pjax's back/forward handling keeps working.

### DIFF
--- a/app/assets/javascripts/errbit.js
+++ b/app/assets/javascripts/errbit.js
@@ -2,7 +2,7 @@
 
 $(function() {
 
-  var currentTab = "summary";
+  var currentTab = window.location.hash.replace(/^#/, '') || "summary";
 
   function init() {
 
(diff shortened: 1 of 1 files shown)


mikelolasagasti (approved):
Non-blocking suggestion: `activateTabbedPanels()` calls `activateTab()`, which already updates the URL with `history.replaceState`. The additional `history.replaceState` in `pjax:end` is therefore redundant and could be removed. `activateTabbedPanels()` itself should remain.

espen:
Thanks. Updated with suggested changes

mikelolasagasti:
@espen can you update the commit to fix the merge conflict?

espen:
let me know if there is anything else to fix

PR #3140, opened 2026-09-04
Cleanup

biow0lf:


mikelolasagasti (changes requested):
Copilot's finding in `app/helpers/apps_helper.rb` remains unfixed.

biow0lf:
> Copilot's finding in `app/helpers/apps_helper.rb` remains unfixed.

Let me show code:

```ruby
  def any_github_repos?
    detect_any_apps_with_attributes if @any_github_repos.nil?

    @any_github_repos
  end
```

This is one from 5 issues. Helpers should not use instance variables. I will handle those issues later in another PR/PRs.

PR #3146, opened 2026-09-08
Keep issue links in the problems table without a configured tracker

bmwiedemann:
The table only linked a problem's issue while its app still had an issue tracker configured, although the problem itself records both the link and the tracker it came from. Show the link whenever the problem carries one, so it stays reachable after the tracker was removed or replaced.


mikelolasagasti (approved):
Makes sense to keep showing existing issue links after the app's issue tracker is modified or removed.

